PR #8, opened 2020-10-19
Adjust headings

gavinorland:
Set headings to be from H1 large downward. Also:

• Minor tweaks to API error message formatting
• Set user's name on dashboard to Keycloak name
• Render main headings as h1, subheadings as captions


shepda (approved):
LGTM

PR #10, opened 2020-10-26
Setup

jennikate:
- hard coded auth and dev URL so we can test while waiting for local to be setup for use
- fixed linting errors
- ejected react-app so could update babel, jest, husky
- updated package.json to latest versions
- updated jest snapshot on changed file

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -58,7 +58,7 @@
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged",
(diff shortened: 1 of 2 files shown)


pixie79:
This needs pulling - you can not hard code urls. This is a public project and a security breach

pixie79 (changes requested):
[Redacted]

gavinorland:
Ejecting CRA is a big change that should perhaps be discussed first, too.

PR #17, opened 2020-11-03
COP-2797 Add 2x logo as per GDS requirements

jennikate:
AC: Footer needs to show logo

Checked GDS and it is looking for the 2x logo. Added this to our images.


jgar123 (approved):
👍

PR #18, opened 2020-11-04
COP-3202 Add route for case page

jennikate:
In this PR

- create a page for 'case'
- add test for case page loads
- add a route to /case
- update test for route


jgar123 (approved):
👍

PR #19, opened 2020-11-04
Update ref data variable to one used by form-api

jennikate:
Confirmed with Anil we should use `referenceDataUrl` and he is updating the secrets as that is what the form api looks for


jgar123 (approved):
👍

jennikate:
just going to close this and open a new one that is signed

PR #21, opened 2020-11-04
Update reference data variable to match form api

jennikate:
Confirmed with Anil, we need to use `referenceDataUrl` as the variable as this is what the form api looks for. He's updated secrets as well


jgar123 (approved):
👍

PR #25, opened 2020-11-05
Revert refdata variable

jennikate:
Confirmed with Mark we should use refData.url
Testing implications on dev once this is merged to see how it will impact existing forms


pixie79 (approved):
lgtm

PR #27, opened 2020-11-06
COP-2826 Add details panel frame

jennikate:
### AC
Default details panel should exist for when there is no case for the businessKey specified

### Updated
- added central section
- added titles and separator
- added results panel
- added non functioning button

### Notes
- The placeholder for businessKey text is visible so we can sign off the design - this panel normally would not show without a businessKey being clicked on/entered

### To test
- Pull and run cop-ui
- Login to cop-ui
- Navigate to the /cases page without error
- You should see the centre section with 4 titles, seperator blue lines, a copy case link button, a dropdown field in case history, an example link in case history


jgar123 (approved):
👍

jennikate:
closing this one and will reopen when worked out why my commits aren't auto signing anymore!

PR #30, opened 2020-11-06
COP-2793 Update header, product name and link to dashboard

jgar123:
### AC
Whitespace around phase banner should be removed and the dashboard url is set to /dashboard

### Updated
- Refactored header/index to prevent the Sign out nav link from overlapping the blue bar below. Removed HeaderContent as replaced with GDS compliant header
- Updated routes/index.jsx to render the dashboard when the route is '/dashboard'. Also added a redirect from '/' to '/dashboard' incase users type the base url with no /dashboard

### To test
- Pull and run cop-ui
- Login to cop-ui
- Navigate to the base url (just '/')
- This should redirect the user to the /dashboard url and render the dashboard


jennikate (approved):
👍

PR #32, opened 2020-11-06
COP-2826 Split details into components

jennikate:
### AC
Default details panel should exist for when there is no case for the businessKey specified

### Updated
- moved sections to their own components
- moved content to translations.json

### Notes
- The h3 on the intro is hardcoded to 'Cases' for now, functionality will be added to make it the businessKey

### To test
- Pull and run cop-ui
- Login to cop-ui
- Navigate to the /cases page without error
- The page should still contain the case actions, history, attachment, metrics sections
- There should be no instances of `pages.cases.xxx`, all should be user friendly words

### DIFF
--- a/cases/CasePage.jsx
+++ b/cases/CasePage.jsx
@@ -23,7 +23,7 @@ const CasePage = () => {
             <input
               spellCheck="false"
               className="govuk-input search__input"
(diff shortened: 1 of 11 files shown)


jgar123 (approved):
👍

PR #35, opened 2020-11-10
COP-3590 Add override scss for header when displaying form

jennikate:
*Bug*

When viewing /forms we were seeing a narrow header, it was using the GDS 1020px version rather than the COP 1400px version. Looked to be due to the DisplayForms component.

*Updated*

Added override css for that component


jgar123 (approved):
👍

PR #39, opened 2020-11-12
COP-3591 Fix header on tasks

jgar123:
### AC
Header is correctly formatted on the /tasks page

### Updated
- Defined the govuk-page-width variable at the top of the TaskList.scss file and set it to 1400px
- Changed fix value in DisplayForm.scss to govuk-page-width variable

### DIFF
--- a/client/src/components/form/DisplayForm.scss
+++ b/client/src/components/form/DisplayForm.scss
@@ -1,3 +1,7 @@
 $govuk-page-width: 1400px;
 
 @import '~govuk-frontend/govuk/all';
+
+.govuk-header__container, .govuk-width-container {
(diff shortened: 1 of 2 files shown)


jennikate:
For reference as to why we chose this path: the pages that display forms call on the Digital Patterns GDS css.
In that, as per GDS guidelines the width is set to 1020px.
For COP we use 1400px.
As the Digital Patterns & GDS css will not be changed, we have added an override here to force this page to always show the 1400 header width.

jennikate (approved):
I've clicked back and forth and refreshed and restarted and I only see the wide header now so I think this has done it!

👍

PR #40, opened 2020-11-12
COP-2819 Add Cases box

jgar123:
### AC
User should be able to click on a cases panel on the dashboard that takes them to the /cases url

### Updated
- Updated translation.json to contain content for cases
- Card title rendering now includes a condition to check if there is both a count and title to display, if so then render both items accordingly. If this this combination is not present then only render the title provided (Forms, Cases, Reports)
- Added default values for Card props that are now not required to be passed down, this is done to allow for the conditional check mentioned above.
- Removed axios call in home/index.jsx that returns the number of forms available to the user, this is does not appear on COP on the dashboard and so has been removed.
- Changed jsx in home/index.jsx to be in accordance with the existing cop's ui, specifically, making the panels be 1/3 of their container and placing each row of panels in an unordered list
- Added tests for the cases panel
- Removed assertions in index.test.jsx that check for the count of forms being rendered on the dashboard. This is not a desired feature of COP.

### Notes
- The `renderTitle` function in the `Card.jsx` file is used in order to achieve identical formatting to COPv1 UI. The main title for each panel is not consistent and depends on whether a count is present. In order to simplify this, I decided to conditionally render 2 types of panel headers each with their respective font size and format.

### To test
*COP UI localhost:*
- Pull and run cop-ui on localhost
- Login to cop-ui
- Click the 'Cases' panel
- You should be navigated to the /cases page without error

*COP UI local:*
- Pull and run cop-ui local with new gradle build
- Login to cop-ui
- Click the 'Cases' panel
- You should be navigated to the /cases page without error

### DIFF
--- a/client/public/locales/en-GB/translation.json
+++ b/client/public/locales/en-GB/translation.json
@@ -67,6 +67,10 @@
         "tasks": {
           "title": "Tasks",
           "footer": "Claim a task assigned to you or your groups"
+        },
+        "cases": {
+          "title": "Cases",
+          "footer": "View active or closed cases"
         }
       }
     },
--- a/client/public/locales/en-GB/translation.json
+++ b/client/public/locales/en-GB/translation.json
@@ -65,7 +65,7 @@
           "footer": "View operational reports"
         },
         "tasks": {
-          "title": "Tasks",
+          "title": "Tasks assigned",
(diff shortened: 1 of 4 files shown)


jennikate (approved):
👍

PR #43, opened 2020-11-16
COP-3613 Add page title for forms

jennikate:
**AC**

Forms should have a page title (based off the `name` in the data, which is set in the related BPMN. This goes above the form's steps.

**Updated**

Added API call to get the `name` and set it as the `<h1>`

**To test**

- click on `record border event` form
- you should see
 - Page Title: `Collect event at border`
 - Form Title: `Record the details of a border event`


jgar123 (approved):
👍

PR #45, opened 2020-11-17
COP-3722 Content

jennikate:
**Updated**
- move object keys to be ordered by common, then pages by alphabetical
- update tasks keys and wording to match what we use in v1
- add group tasks keys and wording

**To test**
- open cop-ui
- click through pages
>- there should be no instances of text looking like `task.list.heading`
>- all titles should be readable
>- task box on dashboard and /tasks should match what is seen in cop v1


jgar123 (approved):
👍

PR #46, opened 2020-11-17
COP-2837 Not found error

jgar123:
### AC
If page is not found, a Not Found page should be rendered and should have a link that routes towards the dashboard

### Updated
- Added a NotFound.jsx component to render when a 404 occurs
- Added barebones test file for Notfound and snapshot
- Added a NotFoundBoundary wrapper to `layout/index.jsx` to catch 404s
- Moved logging capability out of jsx and into it's own reusable function

### Notes 
- The content of the NotFound.jsx component needs reviewing, for now, it closely follows the government template
- This is just one part of the ticket, further logic needs to be added for other error cases
- NotFoundCallback in `layout/index.jsx` was created in order to move chunks of code out of the jsx.

### To test
- Pull and run a build with gradle in cop-ui directory
- Run cop-ui local with new build
- Login to cop-ui
- From the dashboard, after the `/` in the url add `ThisIsNotAPage`
- Repeat this for /forms, /reports, /cases and /tasks
- You should see a Page Not Found message being rendered
- When clicking on the dashboard link provided, you should be taken back to the root url `/` without error
- When clicking on the header navigation links, you should be redirected without error


jennikate (approved):
Discussed & reviewed 👍

jgar123:
A change in how the Page Not Found error was detected and rendered occurred towards the end of this piece of work. In the commit `3eb4d09`, it was handled by the `ErrorBoundary` wrapper which used the `error.status` key to conditionally render the `NotFound` component. This approach worked however, the `ErrorBoundary` component's state is not reset when navigating away from the component when clicking on the `Header` nav links thus not navigating and rendering the correct components (essentially stuck on the Page Not Found screen). `NotFoundBoundary` does not suffer this problem and so a decision was made to move towards this wrapper. 

Having said this, a solution for "resetting" the `ErrorBoundary` component's state will need to be investigated in order to gracefully handle other types of errors we may encounter. For now, `NotFoundBoundary` provides suitable functionality.

PR #50, opened 2020-11-23
COP-3765 Add task filter bar to task list

jgar123:
### AC
Add filter bar to task list

### Updated
- Added a prop to pass down to `your-tasks` and `tasks` routes in `task/routes.js`, it passes a string down to the route to distinguish between task list content
- Created `TaskFilter` component for the task list page

### Notes
- The filter bar is not functional currently, it will be implemented fully later
- Need to change rendered tasks design in a later PR

### To test
- Pull and run cop-ui locally
- Login
- Navigate to `/tasks`:
  - You should see the dropdown selectors without error
  - Select different options on the dropdowns
  - You should not see any changes to the tasks displayed on screen
- Navigate to `/tasks/your-tasks`:
  - You should see the dropdown selectors without error
  - Select different options on the dropdowns
  - You should not see any changes to the tasks displayed on screen

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -44,6 +44,16 @@
     "import/no-extraneous-dependencies": ["error", { "devDependencies": true }],
     "import/prefer-default-export": "off",
     "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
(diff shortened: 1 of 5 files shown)


jennikate (approved):
👍

PR #51, opened 2020-11-24
COP-3765 task list design

jgar123:
### AC
The tasks list page design should be the same as COPv1 UI task list page

### Updated
- Added dayjs package, this is due to momentjs not being maintained anymore and has recommended a switch to a new package (dayjs being a suggestion)
- Refactored TaskList.jsx removing large chunks of jsx and moved them out to separate files (TaskListItem.jsx)
- Added TaskListItem.jsx to handle rendering each individual task
- Added initial tests for TaskListItem.jsx

### Notes
- The functionality is still limited (can't claim task, different grouping, sorting, search etc) as this ticket is for setting up the design, further functionality will be added in later PRs
- Business key is not returned from tasks api call and so the task id is being used as a placeholder for now

### To test
- Pull and run cop locally
- Login
- Navigate to /tasks
- You should see a list of tasks that looks like COPv1 UI
- You should be able to click on a task and it should take you to a single task page

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -4729,6 +4729,11 @@
         }
       }
     },
+    "dayjs": {
+      "version": "1.9.6",
+      "resolved": "https://registry.npmjs.org/dayjs/-/dayjs-1.9.6.tgz",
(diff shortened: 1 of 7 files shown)


jennikate (approved):
👍

PR #52, opened 2020-11-25
COP-2808 Sort tasks by due date/created date/priority

jgar123:
### AC
User is able to sort tasks by due date, created date, and priority

### Updated
- Set the default values for sorting and grouping to be the same as COPv1 UI
- Added `sorting` to json body request to cop-service
- Only trigger api call to cop-service when sorting and search have been used, not grouping
- Changed `TaskFilters.jsx` option values to match the format in the camunda API docs
- Added more tests to `TaskFilters.jsx` 

### Notes
- Group by and search functionality have not been added yet

### To test
- Pull and run cop-ui locally (spin up docker containers etc.)
- Login
- If you have already submitted forms and have tasks then skip the next step
- Submit at least 2 forms (record border event can be submitted pending form changes, this may require making officer details not mandatory in order to complete the form and submit successfully).
- Navigate to `/tasks`
- The tasks should be sorted by oldest due date by default
- Select `Latest due date` from the sort by dropdown
- The tasks should be sorted by latest due date

### DIFF
--- a/client/src/pages/tasks/TasksListPage.jsx
+++ b/client/src/pages/tasks/TasksListPage.jsx
@@ -32,6 +32,11 @@ const TasksListPage = ({ taskType }) => {
     setFilters({ ...filters, [e.target.name]: e.target.value });
   };
 
+  const formatSortByValue = (sortValue) => {
+    const [sortOrder, sortVariable] = sortValue.split('-');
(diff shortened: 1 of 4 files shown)


p3dr0migue1 (approved):
LGTM

PR #53, opened 2020-11-26
COP-2809 Group tasks visually by category/BF ref/priority/assignee

jgar123:
### AC
Users should be able to group tasks by category, bf reference (business key), priority and assignee

### Updated
- Added api call to `/process-instance` to associate a business key with each task and updated `TaskListPage.test.jsx` to reflect the extra call
- Refactored `TaskListPage.test.jsx` slightly moving `mockData` higher in scope so other tests can access it (given more assertions will be added - they will)
- Added assertions to test grouping in `TaskListPage.test.jsx`
- Updated `TaskFilters.jsx` bf reference option value to `businessKey` as this matches the property name in the reponse returned in the `/process-instance` call
- Passed `groupBy` state to `TaskList` and created the task groupings dynamically
- Changed pluralisation of task depending on number of tasks in the rendered groups
- Added a `utils.js` file for potentially reusable functions within tasks i.e. `formatPriority` return correct `priority` label depending on the `priority` value

### Notes
- Will add further tests soon an a separate PR

### To test
*If able to run cop-ui local (docker containers etc.):*
 - Pull and run cop-ui
 - Login
 - Submit at least 2 forms (record border event can be submitted pending form changes, this may require making officer details not mandatory in order to complete the form and submit successfully).
 - Navigate to `/tasks`
 - You should see the tasks grouped by default to category
 - Try each option from the Group by dropdown
 - You should see tasks grouped correctly (currently this is restricted by the type of forms you can submit i.e. grouping by category will always be Border Event if only Record Border Event is the only type of form that can be submitted)

### DIFF
--- a/client/src/pages/tasks/TasksListPage.jsx
+++ b/client/src/pages/tasks/TasksListPage.jsx
@@ -182,7 +182,7 @@ const TasksListPage = ({ taskType }) => {
       </div>
       <div className="govuk-grid-row">
         <div className="govuk-grid-column-full">
-          <TaskList tasks={data.tasks} />
+          <TaskList tasks={data.tasks} groupBy={filters.groupBy} />
           {data.total > data.maxResults && data.tasks.length < data.total ? (
             <ul className="govuk-list">
               <li>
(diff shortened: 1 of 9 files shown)


p3dr0migue1 (approved):
LGTM

PR #54, opened 2020-11-30
COP-2810 Search by task name

jgar123:
### AC
User is able to search for a task by name

### Updated
- Split out `data` state in order keep only directly related data together i.e. maxResults is now a constant as it's value never changed
- Added `filter.search` to request payload to both `tasksResponse` and `taskCountResponse` to allow for search functionality
- Added `taskTypePayload` which decides the type of tasks to search for (either your tasks or your teams tasks)
- Swapped the group name and number of tasks in each group around to match COPv1 UI

### Notes
- API calls on the dashboard need to be revisited for the task count to ensure they are accurate and reflective of their respective task counts

### To test
- Pull and run cop-ui locally (docker containers etc)
- Login
- Submit forms (if able to do so)
- Navigate to /tasks
- Enter the name of the task you want to search
- If the task exists under the name you searched you should see it rendered, otherwise you should not see the task

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -36,6 +36,12 @@
       "rules": {
         "max-len": 0
       }
+    },
+    {
+      "files": ["TasksListPage.jsx"],
+      "rules": {
+        "no-param-reassign": "off"
+      }
     }
   ],
   "rules": {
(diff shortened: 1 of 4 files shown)


gavinorland (approved):
LGTM :) We had a chat and said we will remove any surplus comments later on.

PR #55, opened 2020-12-01
Make context objects available to forms

gavinorland:
This is intended to make all expected context objects and properties available to forms, except `shifthistoryid`. We're aware this is used in some forms and will need to investigate why, because shifts are being dropped. There is one linting exception inline which I will revisit later. We also need to bear in mind for the future that users who have not been through onboarding will not have a `staffid`.


shepda:
I think at some stage we would do well by splitting `client/src/utils/hooks.js` into separate files.

shepda (approved):

PR #57, opened 2020-12-02
COP-2813 Enable pagination on task list pages

jgar123:
### AC
User should be able to navigate through the tasks list with the pagination

### Updated
- Removed `useRef` and and `loadMore` from `TasksListPage` as they aren't required for pagination functionality
- Removed load more tasks assertion from `TasksListPage.test`
- Added `TaskPagination` component to `TasksListPage` that is conditionally rendered depending on the number of tasks returned
- Added `TaskPaginationButton` to allow for conditionally disable buttons
- Added assertions to `TaskPagination.text.jsx`, focused on testing if the buttons were correctly disabled under certain circumstances and vice versa
- Format comments

### To test
- Pull and run cop locally
- Login
- If 21+ tasks are already present then navigate to /tasks, if not, you will need to submit forms until this task number is reached
- *You should see the pagination buttons underneath the task list with '< previous'  and '<< first' being greyed out*
- Click on the 'last >>' button
- *You should be taken to the last page successfully with 'next >'  and 'last >>' being greyed out*
- Click on the '<< first' button
- *You should be taken to the first page successfully '< previous'  and '<< first' being greyed out*
- Next and previous are relative to the amount of tasks you have however they should also be greyed out when at either the first or last page of task results

### DIFF
--- a/client/src/pages/tasks/TasksListPage.jsx
+++ b/client/src/pages/tasks/TasksListPage.jsx
@@ -40,8 +40,10 @@ const TasksListPage = ({ taskType }) => {
     const loadTasks = async () => {
       if (axiosInstance) {
         try {
-          /* taskTypePayload uses the taskType prop to query either the user assigned tasks or 
-          their assigned and group assigned tasks */
+          /*
+           * taskTypePayload uses the taskType prop to query either the user assigned tasks or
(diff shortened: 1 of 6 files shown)


jennikate (approved):
👍

PR #58, opened 2020-12-03
COP-3774 Multistep forms

jennikate:
### AC
User should automatically be shown the next form in a series when relevant

### Updated
- Removed hooks.js file which contained the submit form function as needed to refactor the function and it was easier to do within the form component. We can refactor this out later if it makes sense to
- Added a check for open tasks (camunda indication that there is a form to complete) once current form is submitted
- Added a trigger to load the next form

### Notes
- This feature is still WIP however we are getting read timeouts on the local environment when trying to load the next form, deploying to dev to test if this is a local issue or a code issue. Will revert this PR when testing done

### To test
- Submit a collect event at border form with type Air Passenger
- Submit form
- Expect peopleEaB form to load


jgar123 (approved):
Approved to allow testing on dev

PR #59, opened 2020-12-03
COP-3774 multistep forms

jennikate:
Moved functions back inside UseEffect as the build for master was failing due to them being outside


jgar123 (approved):
Approved for testing on dev

PR #60, opened 2020-12-03
COP-3774 Remove isMounted check, alertContext, add tracking

jennikate:
Removing the isMounted check from the nextStep form loading
Commenting out alertContext for the time being to remove the if statement until we know form is loading correctly
Adding console logs to track data at various states


jgar123 (approved):
Spoke regarding comments and console logs, these are required in order to test certain scenarios in dev

PR #64, opened 2020-12-03
COP-2811 Change dashboard task count requests

jgar123:
### AC
User should see the correct number of tasks assigned to them and their team on the dashboard

### Updated
- Added extra api call to grab the task count for tasks only assigned to the user

### To test
- Pull and run cop locally
- Login
- Submit an intel referral form
- *You should see on the dashboard, the task count for group tasks has increased by 1 whereas the task count for your assigned tasks has stayed the same*


jennikate (approved):
👍

PR #66, opened 2020-12-07
COP-3774 : Enable follow on forms to load

jennikate:
### AC
When a form is submitted, if the next task to be done is assigned to the current user, that task(form) should load immediately.

### Updated
- Added api call to check for tasks and load task page if found
- Added tests

### Notes
- There is now an error showing related to cash/items/firearms seized but it does not seem to be related to this function so will look at it separately

### To test
- Pull and run cop locally
- Login
- Submit a collect-event-at-border form for Air Passengers
- > on submit you should see the People form load
- Submit an intel-referral form
- > on submit you should be returned to the dashboard with a businessKey
- Check group tasks 
- > you should see the Enhance task for this businessKey (you need /devTesting role)
- Submit a Corona virus form
- > you should be returned to the dashboard with a businessKey (there are no further tasks)

### DIFF
--- a/client/src/pages/forms/hooks.js
+++ b/client/src/pages/forms/hooks.js
@@ -1,4 +1,5 @@
 import { useCallback, useContext } from 'react';
+import { useKeycloak } from '@react-keycloak/web';
 import { useTranslation } from 'react-i18next';
 import { useNavigation } from 'react-navi';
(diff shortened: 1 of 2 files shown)


jgar123 (approved):
👍

PR #67, opened 2020-12-08
COP-2811 Change task query parameters

jgar123:
Changed the query parameters for the task list and task count. Before this change, tasks that were assigned to a someone else other than the current user were not visible.


jennikate:
collect-event-at-border tasks are assigned to an individual not a group, so we would not expect to see them in the group list. Raising with Tom C to see if that's an issue if it is, we'll make a card for the feature.

jennikate (approved):
👍

PR #68, opened 2020-12-09
Refactor Power BI work

gavinorland:
This is — unusually — in one commit, because so much of it interlinks.

• Separate various Power BI functions/hooks into their own files so that the code is more modular and allows unit testing
• Move Power BI branch names into constants file
• Remove 'return to reports list' link in favour of new report reload functionality (tweak Sass accordingly)
• Handle user having no team (in this case branch name is not fetched and filters are not set on page change)
• Other more minor refactoring
• Enhance unit tests


shepda (approved):
LGTM

PR #69, opened 2020-12-09
COP-4201 Persist task filters

jgar123:
### AC
The user should be able to alter task filters options and persist upon refreshing the page

### Updated
- Add the `secure-ls` package as used in COPv1 UI for local storage
- Added methods to the `SecureLs` class to keep component code concise
- If `TaskListPage` filters have already been set in local storage, those values are used else use the default value specified
- Passed all filter values to `TaskFilter` component in order to correct update the rendered filter selections
- Refactor `TaskListPage` test

### Notes
- The task filter persistence for group and your tasks are independent of each other

### To test
- Pull and run cop locally
- Login
- Navigate to `/tasks`
- *You should see the task filters set to their defaults 'Oldest due date' and 'Category'*
- Change these filters to other options i.e. 'Highest priority' and 'Assignee'
- *You should see the task list update with the corresponding filter selections*
- Navigate away from the page and navigate back to `/tasks` (or refresh page)
- *You should see most recent filter options selected being the rendered - the task list rendered should also correspond*
- Navigate to /tasks/your-tasks
- *You should see the task filters set to their defaults 'Oldest due date' and 'Category'*
- Change these filters to other options different to those chosen in /tasks i.e. 'Lowest priority' and 'BF reference'
- *You should see the task list update with the corresponding filter selections*
- Navigate away from the page and navigate back to `/tasks/your-tasks` (or refresh page)
- *You should see most recent filter options selected being the rendered - the task list rendered should also correspond*
- Navigate to /tasks
- *The selections made on /tasks/your-tasks should not be the same as /tasks*

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -4391,6 +4391,11 @@
         "randomfill": "^1.0.3"
       }
     },
+    "crypto-js": {
+      "version": "3.3.0",
+      "resolved": "https://registry.npmjs.org/crypto-js/-/crypto-js-3.3.0.tgz",
+      "integrity": "sha512-DIT51nX0dCfKltpRiXV+/TVZq+Qq2NgF4644+K7Ttnla7zEzqc+kjJyiB96BHNyUTBxyjzRcZYpUdZa+QAqi6Q=="
+    },
     "css": {
       "version": "2.2.4",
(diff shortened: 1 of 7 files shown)


jennikate (approved):
👍

PR #70, opened 2020-12-10
COP-2822 Follow on task forms

jennikate:
### AC
If there is a follow-on form from a task, and this form is assigned to the current user, it should display

### Updated
- added follow on form logic to task submit hook
- updated tests for task submit hook
- reviewed why tests were failing for TaskPage:
- - removed the navigation assertion as the navigation now happens in the submit and not from the TaskPage
- - removed click continue button test, discussing with Tom C the expected behaviour and will then implement that

### To Test
_Scenario One: starting with a new form_
- fill out a collect-event-at-border form with seized items, cash, and firearms
- click 'submit'
> you should see the next form load (this is triggered by the form submit hook)
- complete that form
- click submit
> you should see the next form load (this is triggered by the task submit hook)

-Scenario Two: follow on form started from task_
- fill out a collect-event-at-border form (with any detection)
- click submit
- DO NOT submit the next form, instead go to your tasks
- click on your event-at-border next task
- complete that form
- click submit
> you should see the next form load (this is triggered by the task submit hook)
- complete all forms
> you should be returned to tasks with confirmation message

_Scenario Three: no follow on forms from the task_
- submit an intelligence referral form
- go to your group tasks
- click 'claim' against the enhance task
- fill out the enhance task
- click submit
> you should be returned to tasks with a completed message

### DIFF
--- a/client/src/pages/tasks/TaskPage.jsx
+++ b/client/src/pages/tasks/TaskPage.jsx
@@ -19,13 +19,20 @@ const TaskPage = ({ taskId }) => {
   const navigation = useNavigation();
   const [keycloak] = useKeycloak();
   const [submitting, setSubmitting] = useState(false);
-
   const { submitForm } = apiHooks();
   const [task, setTask] = useState({
     isLoading: true,
     data: null,
   });
+
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


jgar123 (approved):
👍

PR #71, opened 2020-12-10
Update accessibility statement

gavinorland:
- Use config property for support URL
- Add changes as in UI V1 (including further edits)


shepda (approved):
LGTM

PR #72, opened 2020-12-10
COP-2822 Showing/Hiding form based on assignee

jennikate:
## AC
If the task assignee and current user are the same, then show the form.
If there is no task assignee, do not show the form.
If there task assignee and current user are NOT the same, do not show the form.

## Notes
- agreed with Tom C that we should remove the 'complete' button as if you can't see the form you should not be completing the task

## Updated
- added check for assignee === current user
- updated state resets
- moved assignee into state
- removed complete button

## To test
- create a task via postman assigned to jennifer.wallace@digital.homeoffice.gov.uk
- create a few more tasks assigned to you or null
- go to your group tasks
- click on the task assigned to Jennifer
>- you should not see a form
- go back
- click on a task assigned to you
>- you should see a form
- go back
- click on a task that is unclaimed
>- you should not see a form
- go back
- claim the task that was assigned to Jennifer
>- you should see a form
- go back
- claim the task that was unclaimed
>- you should see a form

### DIFF
--- a/client/src/pages/tasks/TaskPage.jsx
+++ b/client/src/pages/tasks/TaskPage.jsx
@@ -18,6 +18,7 @@ const TaskPage = ({ taskId }) => {
   const axiosInstance = useAxios();
   const navigation = useNavigation();
   const [keycloak] = useKeycloak();
+  const currentUser = keycloak.tokenParsed.email;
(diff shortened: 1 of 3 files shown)


jgar123 (approved):
👍

PR #73, opened 2020-12-11
COP-4210 Set task page to auto-refresh the task list every 5min

jennikate:
## AC
Task pages should update their task list every 5 minutes

## Updated
- added a 5min setTimeout to loadTasks
- added content to say list will update every 5 min

## To Test
- go to your tasks
- leave page open
- submit a collect event at border task through Postman
- wait 5min
>- you should now see the next EaB task for the businessKey you submitted via postman in the your tasks list (without having refreshed the page yourself)
- go to group tasks
- leave page open
- submit an intel referral task through Postman
- wait 5 min
>- you should now see the an enhance task for the businessKey you submitted via postman in the group task list (without having refreshed the page yourself)

### DIFF
--- a/client/public/locales/en-GB/translation.json
+++ b/client/public/locales/en-GB/translation.json
@@ -185,7 +185,8 @@
         "heading": "{{count}} tasks assigned to your team",
         "status": "Open",
         "title": "Your team's tasks | Central Operations Platform"
-      }
+      },
+      "refresh": "This page auto refreshes every 5 minutes"
     }
   }
 }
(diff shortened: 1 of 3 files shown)


shepda:
In COP V1 we're using websockets to refresh both the task count displayed in the dashboard panel and the tasks displayed in the user's task list. This means that these will get automatically updated as a new task gets assigned to the user. 

This would be a more elegant solution to use than using a timeout to refresh the task list.

jennikate:
> In COP V1 we're using websockets to refresh both the task count displayed in the dashboard panel and the tasks displayed in the user's task list. This means that these will get automatically updated as a new task gets assigned to the user.
> 
> This would be a more elegant solution to use than using a timeout to refresh the task list.

Looked into what Camunda uses for websockets, and it's not super clear. Discussed with Tom C the user need for the refresh and it's not something that is a critical need so closing this PR and will move the ticket into cop-ui v2.1 where we have more time to investigate the need and the best solution for it.

PR #74, opened 2020-12-11
COP-2825 Link to case

jennikate:
## AC
User can click the business key and have the related case open in a new tab

## Updated
Added link

## To test
- click on any task
- click on the businessKey at the top of the screen
> - you should be taken to /cases/<businessKey>


jgar123 (approved):

jgar123 (approved):
👍

PR #76, opened 2020-12-14
COP-3595 Mobile menu not showing

jennikate:
## AC
Mobile menu should appear, and currently is not

## Updated
- updated structure to follow GDS guide for nav
- put service name in the logo space as COP does not use a logo, but just keeps service name on the left
- added toggle functions for nav

## To test
- open cop-ui in a mobile view
- tap on 'menu'
>- you should see the menu items
- tap on 'profile menu item'
> menu should close and take you to profile page
- try tapping menu on various pages
> menu should open/close on tap

### DIFF
--- a/client/src/components/header/index.jsx
+++ b/client/src/components/header/index.jsx
@@ -102,6 +102,30 @@ const Header = () => {
           </div>
         </div>
       </header>
+      <div className="govuk-phase-banner govuk-width-container">
+        <p className="govuk-phase-banner__content">
+          <strong className="govuk-tag govuk-phase-banner__content__tag ">
+            {config.get('uiVersion')}
+          </strong>
+          <span>
+            <strong className="govuk-tag govuk-phase-banner__content__tag ">
+              {config.get('uiEnvironment')}
+            </strong>
+          </span>
+          <span className="govuk-phase-banner__text">
+            {t('header.new-service-1')}{' '}
+            <a
+              className="govuk-link"
+              href={`${config.get('serviceDeskUrl')}/create/54`}
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              {t('header.new-service-2')}
+            </a>{' '}
+            {t('header.new-service-3')}
+          </span>
+        </p>
+      </div>
     </>
   );
 };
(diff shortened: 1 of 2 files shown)


client/src/components/header/index.jsx
```diff
@@ -84,30 +102,6 @@ const Header = () => {
...
-              href={`${config.get('serviceDeskUrl')}/create/54`}
-              target="_blank"
-              rel="noopener noreferrer"
-            >
-              {t('header.new-service-2')}
-            </a>{' '}
-            {t('header.new-service-3')}
-          </span>
-        </p>
-      </div>
```
shepda, line 110: Did you intent to remove the whole phase banner? I'm asking because it's different to what we currently have in V1 in prod and there is no mention of it in the ticket.

If you did intend to remove it then please also remove any config or translation file text that is not used anywhere else.

client/src/components/header/index.jsx
```diff
@@ -8,38 +8,55 @@ import './index.scss';
...
               aria-controls="navigation"
               aria-label="Show or hide navigation menu"
+              onClick={(e) => {
+                toggleMenu(e);
+              }}
             >
               Menu
             </button>
-            <nav className="govuk-grid-column-one-half">
+            <nav id="globalNav">
```
shepda, line 52: It's generally better practice to use classes for styling. It doesn't look like this id is used other than for styling? If not then you might want to consider changing it to a class.

jennikate: I replicated what was done in the original nav file for v2 - which is to use an id to style it, just changed the id name so it was less generic. I have a note to review this for v2.1 when we have more time, to change it to class and ensure it doesn't break anything elsewhere.

client/src/components/header/index.jsx
jennikate: That was unintentional and will be put back.

jennikate: Code added back in now

shepda: :+1:

client/src/components/header/index.jsx
shepda: Ok. I think ideally because this would be a small change it would be better to do now rather than spending more time to revisit, however as long as there is a ticket to revisit then it's fine.

jgar123 (approved):
👍

PR #77, opened 2020-12-14
Fix vulnerabilities

gavinorland:
Fix vulnerabilty warnings by updating packages (including `react-scripts`) and modifying some propTypes and tests as required. This also required moving font assets from the `public` folder into the `src` folder.


shepda (approved):
LGTM

PR #79, opened 2020-12-16
Temporarily add alternative form route

gavinorland:
This is being added so we don't need to edit forms that are shared by both UIs but which link to other forms. I'll add a ticket for us to remove this route and update forms once we've migrated.


shepda (approved):
LGTM

PR #80, opened 2020-12-22
COP-4363 Form submission for certain forms throwing exception on submit

jgar123:
### AC 
User should be able to submit forms successfully without an exception being thrown. NOTE - this PR does not fix 500 responses from form submission.

### Updated
- Moved adding the `form: {}` property to `submissionData from `beforeSubmit` to `onSubmit`. 
- Added `eslintcache` to `gitignore` as the updated `packages` create this automatically but is not needed
- Removed test from `FormPage` as it was dependent on the `beforeSubmit` hook existing.

### Notes
- As mentioned earlier, this PR does not fix 500 responses, only exceptions being thrown

### To test (this will require uploading bpmns and adding forms to local environment)
- Pull and run cop locally
- Login
- Submit intel referral form
- *It should submit successfully*
- Submit collect event at border form
- *It should submit successfully*
- Submit hot debrief form
- *It should submit and return 500*
- Submit assign target form
- *It should submit successfully but have an undefined business key*
- Submit record trade checks
- *It should submit successfully*
- Submit Record use of force
- *It should submit successfully*
- Submit Signage order
- *It should submit successfully*
- Submit Ergonomic assessment
- *It should submit successfully*


jennikate:
Submit Intel referral -- error in Camunda Cockpit `Expression`, this is as expected and there is a ticket to investigate that bug

jennikate (approved):
Checked tasks as some of the tasklist updates are here - they're working as expected
Forms are working as documented above
👍

PR #83, opened 2021-01-06
Remove tag id from dashboard card

jgar123:
Removed the "title" and "count" id from the `renderTitle` function

### To test
- Run cop-ui on localhost
- Login
- Go to dashboard
- *The dashboard should be the same as before*
- *There should be no errors in the console*


jennikate (approved):
👍

PR #84, opened 2021-01-11
COP-4477 Refactor form submission

jennikate:
## AC
Prepare form submission to be able to handle more use cases

## Updated
- consolidated tasks>hooks(submit) & forms>hooks(submit) into a single component file
- added an `id` variable to pass through either `formId` or `taskId` based on the source of the form
- added a `submitPath` variable to pass through the path to be provided to the API call
- updated API call to use `submitPath`/`id` so both tasks and forms can use the same submit component
- updated `FormPage` to use an object to pass through to `submit` hook (this is how `TaskPage` worked)
- add new hooks.js test to test the component file
- updated FormPage and TaskPage tests to use new single component
- removed old files

## To test
- run locally
- make sure your have the relevant bpmns deployed locally with 'devTesting' set as the groups where required (to ensure you can see created tasks)
- submit a single form (e.g. covid compliance checks (without checking the submit another box)
> your form should submit
> you should be taken to the dashboard with a businessKey reference
- submit a form with a follow on (e.g. collect event at border, with items seized)
> your collectEvent form should submit
> the itemsEaB form should be displayed to you immediately
- submit a form that creates a new task but doesn't assign it to you (e.g. submit intelligence referral)
> your submit intelligence form should submit
> you should be taken to the dashboard
> if you view group tasks you should see an enhance intel task for that businessKey


jgar123 (commented):
👍

jgar123 (approved):

PR #85, opened 2021-01-12
Implement Matomo tracking

gavinorland:
This is first steps to getting Matomo tracking up and running across the application. We should see Reports list and pages tracked, and if we do then we'll look at adding other sections.


shepda (approved):
LGTM

PR #87, opened 2021-01-14
COP-2823 Add change priority feature

jgar123:
### AC
When a user is on a single task, they should be able to change the priority of the task and then see the updated version of the task 

### Updated
- Refactored `TaskPage.jsx`, moving the task information at the top of the page into a `TaskPageSummary.jsx` component. 
- Added the `taskUpdateSubmitted` piece of state to reflect when the user has submitted their change to priority - this value is then 'listened' to by the `useEffect` in `TaskPage.jsx`
- `ChangePriority.jsx` added for the change priority functionality
- Added relevant tests and test files for all affected files

### Notes
- An implementation of context was done to avoid prop drilling however, after further consideration it was removed as it didn't fit the convention of how context was used elsewhere in cop-ui. As well as this, only small amounts of prop-drilling was done in order to achieve the desired result.

### To test
- Pull and run cop locally
- Login
- Submit a form (either intel ref or record border event will suffice)
- Access the task raised from the initial submit
- Click the `change` link in brackets next to `Priority`
- *You should see a dropdown with 3 options where the tasks current priority being the default value*
- *You should see a `Change priority` button*
- Select a value from the dropdown different to the one selected i.e. High
- Click `Change priority`
- *You should see the task being refreshed with the priority having being updated correctly*
- Repeat the process from the "Click the `change` link in brackets next to `Priority`" instruction
- *You should see the priority update successfully*

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -42,12 +42,6 @@
       "rules": {
         "no-param-reassign": "off"
       }
-    },
-    {
-      "files": ["TaskPageSummary.jsx"],
-      "rules": {
-        "jsx-a11y/anchor-is-valid": "off"
-      }
     }
   ],
   "rules": {
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -42,6 +42,12 @@
       "rules": {
         "no-param-reassign": "off"
       }
+    },
+    {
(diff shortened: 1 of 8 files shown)


jennikate (approved):
👍

PR #91, opened 2021-01-18
COP-4723 Refactor secure local storage into custom class

jgar123:
### AC
Local storage capability should be the same as before with a different implementation (task filters persist)

### Updates
- Rename SecureLocalStorage to SecureLocalStorageManager
- Place secure-ls logic into custom class (SecureLocalStorageManager)
- Added jest-localstorage-mock for testing local storage
- Add tests for SecureLocalStorageManager

### To test
- Pull and run cop locally
- Login
- Go to group tasks
- Change task filters
- *They should persist on refresh*
- Go to your own tasks
- Change task filters
- *They should persist on refresh*


client/src/utils/SecureLocalStorageManager.js
```diff
@@ -0,0 +1,23 @@
+import SecureLS from 'secure-ls';
+
+export class SecureLocalStorageManager {
```
gavinorland, line 3: This could be the default export, @jgar123?

gavinorland:
I think if this manager class handles setting into localstorage it would be good if it also has the static methods to handle retrieving from localstorage.

client/src/utils/SecureLocalStorageManager.test.js
```diff
@@ -0,0 +1,87 @@
...
+    const result = SecureLocalStorageManager.setInitialState(testKey, 'defaultValue');
+
+    expect(result).toBe(expectedResult);
+  });
+
+  it('should update value associated with key in local storage if key exists', () => {
+    const initialValue = 'foo';
+    const expectedResult = 'bar';
+
+    SecureLocalStorageManager.ls.set(testKey, initialValue);
```
gavinorland, line 34: It feels a bit wrong here for us to be reaching into the class and running methods on its `ls` property (encapsulation - internal implementation might change). Perhaps this should be proxied with a `set` method on the class (which the class itself could then also use in other functions if necessary)?

jennikate (approved):
👍

PR #92, opened 2021-01-19
COP-4280 Update to use ErrorAlert directly and not through AlertContext/Banner

jennikate:
## AC
Form error alerts should not persist

## Updated
DisplayForm to use ErrorAlert rather than AlertBanner

### Scenario One: going back clears the error alert box
 * open collect-event-at-border form
 * click next
 * click next (without selecting any radio buttons)
 > you should see error alerts at the top of the page
 * click 'back'
 > you should no longer see the error alert box

### Scenario Two: going to a previous page of the form clears the error alert box
 * open collect-event-at-border form
 * click next
 * click next without selecting any radio buttons
 > you should see error alerts at the top of the page
 * select a radio button on the first question
 > you should no longer see the error alert for that question at the top of the page
 * select a radio button on the second question
 > you should no longer see the error alerts
 * click next
 * click next without answering any questions
 * > you should see an error alert at the top of the page
 * click previous (button at the bottom of the form)
 * > you should no longer see the error alert

### Scenario Three: submitting a form persists the alert box
 * open the covid-compliance-check form
 * complete the form (do not tick to submit another)
 * submit the form
 > you should be returned to the dashboard with a submit confirmation alert
 * go to another page
 > you should still see the submit confirmation alert
 * click on close confirmation button
 > you should no longer see the submit confirmation alert

### Scenario Four: API error
 * open the hot debrief form (as we know this is currently erroring on COPv2
 * fill out the mandatory questions
 * submit the form
 > you should see an Internal system error alert
 * go back
 > you should still see the internal system error alert
 * refresh the page
 * > you should no longer see the internal system error alert

### DIFF
--- a/client/src/components/form/DisplayForm.jsx
+++ b/client/src/components/form/DisplayForm.jsx
@@ -7,13 +7,13 @@ import PropTypes from 'prop-types';
 import gds from '@digitalpatterns/formio-gds-template';
 import Loader from '@highpoint/react-loader-advanced';
 import { BLACK, WHITE } from 'govuk-colours';
-import { AlertContext } from '../../utils/AlertContext';
 import { TeamContext } from '../../utils/TeamContext';
 import { StaffIdContext } from '../../utils/StaffIdContext';
 import { augmentRequest, interpolate } from '../../utils/formioSupport';
 import Logger from '../../utils/logger';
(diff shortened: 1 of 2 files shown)


jgar123 (approved):
LGTM

PR #95, opened 2021-01-21
COP-4477 Add 'repeat form' functionality

jennikate:
## AC
When a user selects 'enter another' (api key `submitAgain`) and submits, another instance of the form, with a new businessKey, should be loaded

## Updated
- added a `handleOnRepeat` function to FormPage and TaskPage
- updated `hooks.js` submit to check for `handleOnRepeat`
- updated `hooks.jest.js` to test `handleOnRepeat` is called

## Notes
Covid compliance check is the only form to currently use `submitAgain`

## To Test

- open Covid Compliance Check form
- open your network tab
- fill out form and tick enter another
- submit
- note the businessKey for the submission
> a new version of the form should load
- fill out the form and do NOT tick enter another
- submit
- note the businessKey for this submission
> the two businessKeys should be different
> you should be taken to the dashboard with a confirmation box

### DIFF
--- a/client/src/components/form/hooks.js
+++ b/client/src/components/form/hooks.js
@@ -14,7 +14,7 @@ export default () => {
   const currentUser = keycloak.tokenParsed.email;
 
   const submitForm = useCallback(
-    ({ submission, form, id, businessKey, handleOnFailure, submitPath }) => {
+    ({ submission, form, id, businessKey, handleOnFailure, handleOnRepeat, submitPath }) => {
       if (form) {
         const variables = {
           [form.name]: {
@@ -40,6 +40,8 @@ export default () => {
                 // We can only ever open one task in this manner and so always take the first available
                 if (response.data.length > 0 && response.data[0].assignee === currentUser) {
                   navigation.navigate(`/tasks/${response.data[0].id}`);
+                } else if (submission.data.submitAgain === true) {
+                  handleOnRepeat();
                 } else {
                   setAlertContext({
                     type: 'form-submission',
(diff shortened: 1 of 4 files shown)


jgar123 (approved):
LGTM

shepda:
This functionality shouldn't need any UI changes as it's all handled in the workflow

jgar123:
Submitting a Covid Compliance form in cop-ui dev at the moment does not reload the same form when you select 'Enter another spot check' - it takes the user to the dashboard which isn't the same behaviour as v1

jennikate:
As Jonny said ^
The cop-ui code was accepting the response from Camunda and automatically loading the dashboard for ALL forms, regardless of whether the BPMN indicated another should be loaded. Hence this update. We had to make adjustments for collect-event-at-the-border & similar forms where the BPMNs load follow on forms as well.

shepda:
The UI code should be kept very general so we shouldn't be putting workflow specific code in the UI.

This functionality does not have any specific code in the UI in V1 so we should work out why this is not working in V2 and how it is different to V1 and then look at the best way to handle it, not putting quick fixes into the UI code.

jennikate:
v1's custom API returns 
```
  "shouldSubmitAgain": {
        "type": "Boolean",
        "value": false,
        "valueInfo": {}
      }
```
as part of the submit response (processInstance).

I have opened a ticket to investigate what Camunda returns us on the submit-form POST to see how we can identify there is a new task that should be loaded immediately, when that task has a new businessKey (as opposed to a new task that should be opened immediately for the same businessKey or a new task that should not be opened immediately)

PR #96, opened 2021-01-22
Fix name of shiftDetailsContext object

gavinorland:
This fixes the issue of some components not being properly pre-filled in, for example, the COVID form.


shepda (approved):
LGTM

PR #101, opened 2021-01-28
COP-2827+COP-2829: Cases search functionality and history panel

ajames14:
**AC**
- replicate the behaviour from COP-v1 (cop-private-workflow-tasklist) for the cases page.

**Description**
- This branch adds the case search functionality - a user can search a business key or keyword and search results are displayed in left hand panel. 
- When the user selects a case link the details panel appears. So far, the only sections populated are the heading and the Case History section. 

**To test:** 
- Pull branch and run development server against the Dev environment (change proxy in package json). This is because the workflow-service latest build is still not working and you need to have the latest version of the API to get this to work. 
- Navigate to http://localhost:3000/cases and input a search. Results will appear on the left hand panel if there are hits. Select a case and the details panel will show with the Case history showing the process instances.
- "Show Details" form snapshot on Case History not implemented yet.
- npm run test to see tests for CasePage, CaseResults panel, Case History.

**NB:** 
- Cases API does not always bring back the correct "Number of cases found" - This is the same on V1.
- If a search has +20 hits the API paginates results. However, this doesn't seem to always work correctly, e.g. search keyword: abigail - it will say there are 40 results, over 2 pages, but "page 1" only has 6 results and the next has 14.

### DIFF
--- a/client/src/pages/cases/CasesResultsPanel.jsx
+++ b/client/src/pages/cases/CasesResultsPanel.jsx
@@ -1,24 +1,15 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import PropTypes from 'prop-types';
-import _ from 'lodash';
 
-const CaseResultsPanel = ({
-  caseSearchResults,
-  caseArray,
-  getCaseDetails,
-  loadMoreCases,
-  nextUrl,
-}) => {
+const CaseResultsPanel = ({ totalElements, caseArray, getCaseDetails, loadMoreCases, nextUrl }) => {
   const { t } = useTranslation();
-  // Check if caseSearchResults.links has property 'next'
-  const hasMoreData = _.has(caseSearchResults._links, 'next');
 
   return (
     <>
       <h2 className="govuk-heading-m">{t('pages.cases.results-panel.title')}</h2>
       <p className="govuk-body govuk-!-margin-bottom-1">{t('pages.cases.results-panel.caption')}</p>
-      <p className="govuk-body govuk-!-font-weight-bold">{caseSearchResults.page.totalElements}</p>
+      <p className="govuk-body govuk-!-font-weight-bold">{totalElements}</p>
       <ul className="govuk-list">
         {caseArray &&
           caseArray.map((item) => {
@@ -39,7 +30,7 @@ const CaseResultsPanel = ({
             );
           })}
       </ul>
-      {hasMoreData ? (
+      {nextUrl ? (
         <button
           type="button"
           className="govuk-button"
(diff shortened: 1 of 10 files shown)


client/src/pages/cases/components/CaseHistory.jsx
```diff
@@ -25,17 +42,83 @@ const CaseHistory = () => {
...
+                        </div>
+                      </div>
+                    </div>
+                  </div>
+                  {processInstance.formReferences.length !== 0 && (
+                    <FormDetails formReferences={processInstance.formReferences} />
+                  )}
+                  {processInstance.formReferences.length === 0 && (
+                    <h4 className="govuk-heading-s">No forms available</h4>
+                  )}
```
gavinorland, line 109: Perhaps this could be:

```
{ processInstance.formReferences.length ? <FormDetails formReferences={processInstance.formReferences} />
: <h4 className="govuk-heading-s">No forms available</h4> }
```

client/src/pages/cases/CasesResultsPanel.jsx
```diff
@@ -1,16 +1,71 @@
...
   );
 };
 
+CaseResultsPanel.defaultProps = {
+  caseArray: null,
+  caseSearchResults: null,
+};
+
+CaseResultsPanel.propTypes = {
+  caseSearchResults: PropTypes.shape(PropTypes.object),
```
gavinorland, line 64: This will result in a console warning - try:

`caseSearchResults: PropTypes.object,`

client/src/pages/cases/CasesResultsPanel.jsx
```diff
@@ -1,16 +1,71 @@
...
+  caseArray: PropTypes.arrayOf(PropTypes.object),
+  getCaseDetails: PropTypes.func.isRequired,
+  loadMoreCases: PropTypes.func.isRequired,
+  nextUrl: PropTypes.string.isRequired,
```
gavinorland, line 68: This will have a problem with the `null` that is set in CasePage L32.

jgar123 (approved):
LGTM, tests will be added later for components when they have functionality

client/src/pages/cases/CasesResultsPanel.jsx
ajames14: Thanks - I've updated the props to avoid these issues.

client/src/pages/cases/components/CaseHistory.jsx
ajames14: Done!

jennikate (approved):
👍

jgar123 (approved):
LGTM

PR #107, opened 2021-02-02
COP-4634 Add css for grey box on covid compliance form

jennikate:
## AC
Replicate v1 styling for v2 forms

## Updated
Added css to style the grey box to the Forms.scss as it is only used on the one form

## Notes
v1 used Bootstrap to style this, we do not use Bootstrap in v2 and this style isn't a GDS style so have implemented the simplest styling possible onto the FormBuilder classes that exist


jgar123 (approved):
LGTM

PR #108, opened 2021-02-02
COP-4634 add styling for case header width

jennikate:
## AC
COP uses a 1400px header width, make cases use the same

## Updated
Added width setting to cases scss file

## Notes
The standard GDS width style is 900px but we use 1400px on COP. The DP GDS stylesheet keeps to the 900px width so we need to reset that in stylesheets to ensure every page keeps our 1400px width.


jgar123 (approved):
LGTM

PR #109, opened 2021-02-02
Set document title from process name

gavinorland:
Also refactor code and tests somewhat: improve destructuring, rename functions, remove conditionals, remove unnecessary "then" callback.


shepda (approved):
LGTM

PR #110, opened 2021-02-03
COP-2830: Add case history order by start date and other case iterations

ajames14:
**Description:** 
This branch adds the functionality to order the case history by process start date. It also makes a few changes to the cases page including moving text to translate file and using it(should...) format for tests.

**To test:** 
- pull the branch and run locally in development mode, making sure to point the proxy at the dev environment. 
- navigate to the cases page (http://localhost:3000/cases) and search for a case e.g. DEV-20210129-952 which has multiple process instances. 
- Check the order of the case history by opening the accordion elements and seeing the start dates. When case history first loads it should have the "latest" instance first, descending.
- Change the order in the "Order by" select box to "Earliest..." and this should correctly reorder. 
- CaseHistory.test.jsx has tests to check rendering in correct order.


jgar123 (approved):
LGTM

PR #111, opened 2021-02-08
COP-2832 Show case attachments

jgar123:
### AC
User should be able to see case attachments and successfully download them if present

### Updated
- Added functionality to `CaseAttachments` component
- Added relevant tests for `CaseAttachments` component
- Moved content out of `CaseAttachments` into `translation.json`

### To test
- Pull and run cop locally setting the `proxy` in `package.json` to cop-ui dev

SCENARIO 1: 
ON COP-UI DEV:
- Submit `Record border event` form opting to attach file
- Move through to attachment form and attach file with valid file type (keep note of the file you attached)
- Submit form and make note of the business key

ON LOCAL:

- Navigate to cases page and search business key
- Select business key from list
- Scroll down to Case Attachments and click on `View attachments`
- *You should see the file you attached appear in the list*
- Click on the file name
- *Your file should download successfully*

SCENARIO 2: 
ON COP-UI DEV

- Submit `Record border event` form opting NOT to attach file
- Submit form and make note of the business key

ON LOCAL:
- Navigate to cases page and search business key
- Select business key from list
- Scroll down to Case Attachments and click on `View attachments`
- *You should see `No attachments associated with case` appear in the table*

### DIFF
--- a/client/public/locales/en-GB/translation.json
+++ b/client/public/locales/en-GB/translation.json
@@ -68,7 +68,15 @@
           "heading": "Case actions"
         },
         "case-attachments": {
-          "heading": "Case attachments"
+          "heading": "Case attachments",
+          "subheading": "View attachments",
+          "loading-attachments": "Loading attachments...",
+          "table": {
+            "header-1": "Name",
+            "header-2": "Uploaded on",
+            "header-3": "Uploaded by",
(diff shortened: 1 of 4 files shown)


jennikate (approved):
👍

PR #113, opened 2021-02-10
COP-5132 Refactor DisplayForm, remove processContext and taskContext from req payload

jgar123:
### AC
The user should be able to submit forms successfully without receiving a 413 error (payload too large).

### Updated
- Removed `submissionData` from state in `DisplayForm`. The `Form` component has its own `data` object that is updated with each form change and then subsequently validated in the `onChange` handler. This means that the `data` object can be used for `onSubmit` as opposed to the `submissionData` state - the data used in the payload is referred to as `submissionData` but does not refer to state.
- Added `beforeSubmit` hook
- Moved adding of the `form` property to the payload in the `beforeSubmit` hook
- Added removing of context objects (`processContext` and `taskContext`) to the `beforeSubmit` hook

### Notes
- Testing will need to be done in dev to access process history and double check the payloads in the process do NOT contain `processContext` and `taskContext`

### To test
- Pull and run cop locally
- Login
- Open dev tools
- Submit `Record Border Event` form (whilst filling out, purposefully try and submit invalid data or miss mandatory fields - checking errors are rendering as they should)
- Make note of the business key
- Click the `Network` tab and locate and click on the `submit-form` request made
- In the `Request Payload` section, click `view sourced` and copy into an empty file
- Search for the word `processContext`
- *You should NOT find results for `processContext`*
- Search for the word `taskContext`
- *You should NOT find results for `taskContext`*
- Fill out the next form that is loaded and submit
- Click the `Network` tab and locate and click on the second `submit-form` request made
- In the `Request Payload` section, click `view sourced` and copy into an empty file
- Search for the word `processContext`
- *You should NOT find results for `processContext`*
- Search for the word `taskContext`
- *You should NOT find results for `taskContext`*
- Locate the process in cockpit under the business key generated
- Click the `variables` tab
- *There should be no mention of `processContext` or `taskContext` in the json of any of the variables*

### DIFF
--- a/client/src/components/form/DisplayForm.jsx
+++ b/client/src/components/form/DisplayForm.jsx
@@ -228,13 +228,13 @@ const DisplayForm = ({
           setErrorAlert(null);
         }}
         submission={augmentedSubmission}
-        onSubmit={(data) => {
+        onSubmit={(submissionData) => {
           setTime({
             ...time,
             end: new Date(),
             submitted: true,
           });
-          handleOnSubmit(data);
+          handleOnSubmit(submissionData);
         }}
         onChange={(data) => {
           if (formRef.current) {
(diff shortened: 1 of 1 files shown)


jennikate (approved):
LGTM 👍

PR #114, opened 2021-02-10
COP-2834: View Case metrics 

ajames14:
**Description**
This branch adds the functionality to view case metrics when a case is selected. It also updates some of the tests to allow for this change. 

**To test:**
- Pull branch and run locally in development mode, making sure you are using the cop-dev proxy in the package.json.
- Search for any case and select from the list. 
- Under Case metrics drop down you should find the relevant information about the case.


jgar123 (approved):
LGTM

PR #124, opened 2021-02-22
COP-2828: Add Case Actions

ajames14:
**Description**
This branch adds Case Actions functionality and associated tests. It also refactors some of the other case tests to have the data in json fixture files rather than in the main test script. 

**Acceptance criteria**
Replicate V1 case actions functionality

**To test:** 
- run locally and navigate to the Cases page. 
- search for a Case that has multiple case actions e.g. http://localhost:3000/cases/DEV-20210215-2666
- within the Case Actions accordion, you should see the tabs populated with the actions and the first tab showing the relevant form. 
- switching between tabs should render different forms
- submit a form for one of the actions. You should see a confirmation message where the form used to be that holds for 5 seconds. 
- After 5 seconds, the same Case details panel will load, displaying the case action you have just submitted in the Case History below. 

NB: 
- there is still an issue with Record Actions being "re-submittable" as many times as you like. On v1 you can only submit this once as the API throws a 409 for duplicate submissions. 
-**The Generate Case PDF form has a url hard coded into the form that is only applicable to V1. So rather than changing the form, which would break V1, we have added a temporary patch to the formioSupport.js file, in the interpolate form function. This is overwriting the url with the V2 end point**


jennikate (approved):
👍

PR #125, opened 2021-02-25
COP-3793 Remove existingSubmission as it has become redundant

jennikate:
## AC

While investigating how to stop form fields pre-populating unexpectedly (no localStorage for that form, but form is populated with data), tracked down that it was coming from existingSubmission being set.

However, digging into that showed that existingSubmission had become redundant. We believed it was passing data to prepopulate a task, but looking at the way we now add context to tasks based on businessKey/processInstance/taskContext we no longer need a variable to pass this data down, we form it on the task as the form creates.

So, remove existingSubmission and it's associated formSubmission.

## To test

- submit a collect-event-at-border form
- > you should see the follow on form as normal
- submit the follow on form
- > you should be able to complete the event at border in a single form session

- submit a covid-compliance-check form with submit again ticked
- > you should be able to submit the form 
- > you should see a new instance of the form without saved data

- open your tasks
- select a journeyEaB task
- select a mode of transport
- close that task
- open a different journeyEaB task
- > the mode of transport SHOULD NOT be prefilled
- submit this task
- > you should be able to complete the submission

### DIFF
--- a/client/src/components/form/DisplayForm.jsx
+++ b/client/src/components/form/DisplayForm.jsx
@@ -18,14 +18,7 @@ import './DisplayForm.scss';
 
 Formio.use(gds);
 
-const DisplayForm = ({
-  form,
-  handleOnCancel,
-  handleOnSubmit,
-  existingSubmission,
-  interpolateContext,
-  submitting,
-}) => {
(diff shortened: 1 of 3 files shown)


p3dr0migue1 (commented):
LGTM, I have not tested it

jennikate:
Closing without merging as this is not working exactly as expected

PR #126, opened 2021-02-25
COP-3973 Persist partially filled forms

jennikate:
## AC

On forms, if you refresh, any fields you've already filled in will persist so you don't need to start again.

## Updated

- index.js to clear all form related storage when you return home (to Dashboard)
- displayForm to store form details on Change
- displayForm to retrieve details if they exist on pageLoad
- hooks.js, FormPage, and displayForm to handle repeatForms where we do not want fields to persist on submit, but where user will not have returned home between form loads

## Notes
There is a bug in Scenario 6 below that exists on v1, and v2dev. It is not related to localStorage so will be investigated in another ticket.

#### Scenario 1

- start new form fresh
- enter some fields
- refresh page

> answers should persist

#### Scenario 2

- start new form fresh
- enter some fields
- go to edit my profile
- return to form via back button

> answers should persist

#### Scenario 3

- start new form fresh
- enter some fields
- go to dashboard

> answers should NOT persist

#### Scenario 4

- start new form fresh
- enter some fields
- complete the form
- submit the form
- return to form

> answers should not persist

#### Scenario 5

- open a task
- enter some fields
- refresh page

> answers should persist

#### ~~Scenario 6~~

- ~~open a task of the same type as in scenario 5~~

> ~~answers should NOT persist~~
See notes, this scenario fails due to an unrelated bug.

#### Scenario 7

- open covid-compliance-form
- complete the form including selecting submit another

> answers should NOT persist

### DIFF
--- a/client/src/pages/home/index.jsx
+++ b/client/src/pages/home/index.jsx
@@ -6,6 +6,7 @@ import axios from 'axios';
 import { useMatomo } from '@datapunt/matomo-tracker-react';
 import Card from './components/Card';
 import { useIsMounted, useAxios } from '../../utils/hooks';
+import SecureLocalStorageManager from '../../utils/SecureLocalStorageManager';
 
 const Home = () => {
   const { t } = useTranslation();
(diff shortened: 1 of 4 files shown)


client/src/pages/home/index.jsx
```diff
@@ -29,6 +30,24 @@ const Home = () => {
...
+  useEffect(() => {
+    const removeArray = [];
+    // eslint-disable-next-line no-plusplus
+    for (let i = 0; i < localStorage.length; i++) {
+      if (localStorage.key(i).indexOf('form') !== -1) {
+        removeArray.push(localStorage.key(i));
+      }
+    }
+    // eslint-disable-next-line no-plusplus
+    for (let i = 0; i < removeArray.length; i++) {
```
p3dr0migue1, line 46: you could probably do:
```js
for (const item of removeArray) {
    SecureLocalStorageManager.remove(item);
}
```

jennikate: Linter failed on `for of`, on investigation AirBnB recommend using `forEach` rather than `for of`, so I've converted to `forEach`

jgar123 (approved):
LGTM

PR #129, opened 2021-03-01
Revert augmentedSubmission

jennikate:
Temporary fix as discovered contexts is being lost at form creation. This fix breaks localStorage but we will continue working on that

This reverts the augmentedSubmission object to being created as it was originally, and removes the useEffect that was retrieving from localStorage when available


jgar123 (approved):
LGTM

PR #131, opened 2021-03-01
COP-5519: Differing data types in the shiftDetailsContext object across v1 and v2

ajames14:
# Description
This branch changes the number values in the shiftDetailsContext to strings, to be in line with the data types in v1. See ticket for further details and comparison with v1: https://support.cop.homeoffice.gov.uk/browse/COP-5519. 

In v1, the shiftDetailsContext is defined through the submission of a shift form, which is why the data is in string format. In v2, we are building the context through a call to ref data and using keycloak, so the numbers are type number. To avoid any issues with potential type errors, the hook setting team context has been updated along with the location ID in DisplayForm and CaseActions. 

To test: 
- submit a form either on the forms page or in case actions. 
- check the payload or look in cockpit at the submission data. Find the shiftDetailsContext. You will see that the values which are numerical are string type e.g.
```
      "ministryid": "1",
      "departmentid": "1",
      "directorateid": "2",
```
- submit a form on v1 and check the submission data in cockpit. The shiftDetailsContext will be of the same type as on v2.

### DIFF
--- a/client/src/components/form/DisplayForm.jsx
+++ b/client/src/components/form/DisplayForm.jsx
@@ -111,7 +111,7 @@ const DisplayForm = ({
       },
       shiftDetailsContext: {
         email,
-        locationid: defaultlocationid,
+        locationid: String(defaultlocationid),
         phone,
         roles,
         team,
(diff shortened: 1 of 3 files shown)


jgar123 (approved):
LGTM

PR #139, opened 2021-03-04
Update logging to see if we can get more information

jennikate:
We are receiving an error from the Zuul proxying when we upload large files.
Updating logging to see if we can get more information from it
We can't upload files locally due to auth issues, so need to test this on dev


jgar123 (approved):
Solely for testing purposes. LGTM

PR #140, opened 2021-03-04
Add zuul host timeout to test if timeouts are causing 500 errors

jennikate:
Testing while after a file upload we receive a 500 error - cannot test file uploading locally 

This is setting timeouts on the zuul proxying to test if that is the problem, set to 2min for test

### DIFF
--- a/server/src/main/resources/application.yml
+++ b/server/src/main/resources/application.yml
@@ -32,6 +32,9 @@ server:
 
 
 zuul:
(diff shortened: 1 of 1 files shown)


jgar123 (approved):
For testing purposes. LGTM

jennikate:
This has resolved the issue and we've decided to leave it at 2min

PR #141, opened 2021-03-04
Return logging to original form now testing is complete

jennikate:
Return logging to the nicely formatted version


jgar123 (approved):
LGTM

PR #142, opened 2021-03-04
COP-5666: Add copy case link button functionality

ajames14:
Description
- Adds functionality to the "Copy case link" button on Cases details panel. 

To test: 
- run locally and go to the Cases page. 
- search for any case and click on one. 
- In the top panel of the Case details panel is the copy button. Click this and then test it has copied to your clipboard.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -101,7 +101,7 @@
       "npm run lint"
     ]
   },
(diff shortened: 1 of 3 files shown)


jgar123 (approved):
LGTM

PR #143, opened 2021-03-04
COP-5468 Modify task requests for homepage and user tasks

jgar123:
### AC
When a user submits a line manager request, they should see a task appear in their own user task list

### Updated
- Changed `assignee` to `involvedUser` in the camunda API calls for both the users tasks and the dashboard count. Assignee only picked up tasks with the current user assigned to it whereas `involvedUser` includes user associated tasks - not just assignee related

### To test
- Pull and run cop locally pointing the proxy to the dev environment
- Login
- Navigate to `/submit-a-form/set-or-change-line-manager`
- Working with another user, input that user email address and submit form
- *The form should submit successfully*
- *The other user should receive an email in regards to this event*
- Ask the other user to navigate to their `tasks/your-tasks`
- *They should see the line manager event in their list of tasks associated with them*
- Ask the other user to navigate to `/tasks`
- *They should NOT see the line manager event in the list of tasks*

### DIFF
--- a/client/src/pages/home/index.jsx
+++ b/client/src/pages/home/index.jsx
@@ -84,7 +84,7 @@ const Home = () => {
         data: {
           orQueries: [
             {
-              assignee: keycloak.tokenParsed.email,
+              involvedUser: keycloak.tokenParsed.email,
             },
           ],
         },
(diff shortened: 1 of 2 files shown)


jennikate (approved):
LGTM 👍

PR #144, opened 2021-03-05
COP-5475 Add context to hidden component for case actions

jgar123:
### AC
When a user submits a case action, all context related properties should be resolved correctly.

### Updates
- Passed `submission` prop with the `contexts` object down to the `Form` component
- As a result of passing `submission`, context now needs to be removed in the `beforeSubmit`
- Removed `showSubmissionConfirmation` in `setTimeout` to avoid state updates after unmounting

### Notes
- With the downgrade of `formiojs`, bugs in the `Form` component are not present i.e. allowing us to pass a `submission` prop down without the object being passed down overwriting key values unexpectedly.

### To test
- Pull and run cop locally pointing proxy to dev environment
- Login
- Submit form that generates multiple case actions
- Go to case that has been created
- Open the network tab in dev tools
- Submit a `Record action` form
- Assess the payload, make sure to check there are no templates (`{{ VALUE IN HERE }}`) in the payload and that they have all been resolved correctly
- Check the payload in cockpit to ensure the payload is still as expected

### DIFF
--- a/client/src/pages/cases/components/CaseAction.jsx
+++ b/client/src/pages/cases/components/CaseAction.jsx
@@ -23,17 +23,14 @@ const CaseAction = ({
   const isMounted = useIsMounted();
   const [submitting, setSubmitting] = useState(false);
   const [submissionConfirmation, showSubmissionConfirmation] = useState(false);
-  const [hasFormChanged, setHasFormChanged] = useState(false);
-  const host = `${window.location.protocol}//${window.location.hostname}${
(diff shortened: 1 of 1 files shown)


p3dr0migue1 (approved):

p3dr0migue1 (approved):
Code looks good (didn't tested it)

ajames14 (approved):

jennikate:
Payload looks good 👍

PR #145, opened 2021-03-05
COP-5557 Refactor TaskPage

jennikate:
## AC
Understand & fix `existingSubmission` as it is prefilling data on forms in unexpected ways

## Investigated
`TaskPage` was always passing through `existingSubmission` as `undefined`. Determined that based on the current code, it would always be undefined

## Updated
Removed code that was resulting in undefined
Updated get call to simplify it to only retrieving form information and determining whether to display form or not

## Notes
The assignee label breaks with this PR, the logic for which name to show is being moved and will be added in the next PR

## To test

Scenario 1 : follow on forms
- start a collect-event-at-border form
- click submit
- > you should be taken to the next followon form
- complete that form and click submit
- > you should be taken to the next followon form

Scenario 2
- fill out and submit an intel referral form
- go to team tasks
- > you should be able to find the enhance task for the businessKey you just submitted
- open the task (do not claim it)
- > you should not see a form
- go back
- claim the task
- > you should see a form

### DIFF
--- a/client/src/pages/tasks/TaskPage.jsx
+++ b/client/src/pages/tasks/TaskPage.jsx
@@ -4,7 +4,6 @@ import axios from 'axios';
 import { useTranslation } from 'react-i18next';
 import { useKeycloak } from '@react-keycloak/web';
 import { useNavigation } from 'react-navi';
-import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


jgar123 (approved):
LGTM

PR #147, opened 2021-03-08
COP-5557 Extract userContexts to a separate file

jennikate:
## AC
Understand & fix existingSubmission as it is prefilling data on forms in unexpected ways

## Investigating
While investigating how user contexts are created I found some repetition and this PR addresses that.

## Updated
userContexts related code refactored out of DisplayForms into it's own file and refactored to reduce repetition

## To Test
- Submit a form
- Contexts should exist as before


jennikate:
Closing as this is WIP

jennikate:
Updated and ready for review

jennikate:
After better estimation of the extent of work needed to be investigated/tested/refactored to fix the bug this was addressing, we have decided to move the work to v2.1

The bug exists in v1 already.

PR #148, opened 2021-03-09
COP-5557 - Revert the refactoring of TaskPage (keep assignee changes)

jennikate:
This PR reverts most of the refactoring from https://github.com/UKHomeOffice/cop-ui/pull/145

This is because we've discovered that to fix the bug that PR 145 was a first piece of, is much larger than hoped and so it has been moved to v2.1.

We're retaining the change to where the `assigneeText` is set as that resolved a different issue.

## To Test

- check collect-event-at-border
- > you should have follow on forms load correctly
- > you should be able to add attachments
- > you should be able to submit the final form
- > you should be able to see it completed in Cockpit

- check submit intel referral
- > you should be able to complete the form
- > you should see an enhance task in group tasks
- > you should be able to claim and complete that enhance task
- > you should be able to see it completed in Cockpit

- check NS referral
- > you should see the mode & port filter dropdowns populating

- check covid compliance
- > you should be able to tick submit another and see a clean version of the form on submit
- > you should be able to submit a form without ticking submit another
- > you should be able to see both in cockpit

### DIFF
--- a/client/src/pages/tasks/TaskPage.jsx
+++ b/client/src/pages/tasks/TaskPage.jsx
@@ -4,62 +4,114 @@ import axios from 'axios';
 import { useTranslation } from 'react-i18next';
 import { useKeycloak } from '@react-keycloak/web';
 import { useNavigation } from 'react-navi';
+import _ from 'lodash';
 import { useMatomo } from '@datapunt/matomo-tracker-react';
-import { useAxios } from '../../utils/hooks';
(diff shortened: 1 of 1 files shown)


jgar123 (approved):
LGTM, off the back of this PR we have discovered existingSubmission is appending form data unexpectedly in the payload. We have a ticket to fix this issue

PR #152, opened 2021-03-11
COP-3825 Remove outdated package

jennikate:
## AC

Resolve warnings of componentWillMount & similar

## Updated

Determined one source of these was `react-loader-advanced` package which looks to be no longer maintained (last update over 2yrs ago)

It was only used on the `DisplayForm` component to show the `ApplicationSpinner` as `submitting`

I've removed the `Loader` component and replaced it with the ternary that is used elsewhere to determine whether to show spinner or not. The ternary return statement is marked to be reviewed and potentially refactored at a later date.

## To test

- pull this branch
- remove `package-lock` & `node-modules`
- from the client folder run `npm i`
- open a form
- submit the form
- > you should see a spinner with the text 'submitting form'

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -3488,11 +3488,6 @@
         "readable-stream": "^2.0.6"
       },
       "dependencies": {
-        "isarray": {
-          "version": "1.0.0",
-          "resolved": "https://registry.npmjs.org/isarray/-/isarray-1.0.0.tgz",
-          "integrity": "sha1-u5NdSFgsuhaMBoNJV6VKPgcSTxE="
-        },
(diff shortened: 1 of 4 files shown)


ajames14 (approved):
LGTM :)

ajames14 (approved):

jennikate:
Discovered this was impacting the persistence of data incorrectly on the repeat forms, when debugging that it looks like something is persisting the form answers somewhere other than localStorage and setting them back into localStorage and populating the form unexpectedly.

Will review how augmentedSubmission & localStorage work before returning to change how we trigger the applicationSpinner

PR #153, opened 2021-03-11
Force use of formiojs version 4.9.19

jennikate:
Due to later versions of formiojs causing errors (ticket open to investigate and safely update), we want to ensure the app only uses 4.9.19 for now


jgar123 (approved):
LGTM

PR #158, opened 2021-03-16
Strip quote marks from Case Search input

ajames14:
### Description

This branch updates the search functionality for Cases by stripping any " symbols from the search string. This is because the v2 Cases API throws a 500 if you search using a search term with unclosed double quotes e.g. "COP-123 ... 

On v1, users are guided to search using double quotes as this is what the old engine required. No quotes are required on v2 to return the correct cases but users will likely be used to searching in this way, so here we add in logic to strip out all " from the search input to avoid returning 500 errors. 

Also updated the hint to remove reference to searching with quote marks. 

### To test:
- Run locally and point proxy to dev. 
- Navigate to Cases page and try to search for a case with double quotes in. You should see no errors.

### DIFF
--- a/client/public/locales/en-GB/translation.json
+++ b/client/public/locales/en-GB/translation.json
@@ -54,7 +54,7 @@
       "title": "Cases",
       "caption": "Case view",
       "heading": "Cases",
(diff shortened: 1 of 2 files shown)


jgar123 (approved):
LGTM

PR #160, opened 2021-03-19
Use customValidation instead of beforeSubmit hook

gavinorland:
If we try to perform validation that depends on certain context objects which are removed when the form is submitted, we run into a problem when the `beforeSubmit` hook is used in that they are removed before the validation occurs. If we instead use the `customValidation` hook then these objects are removed _after_ validation occurs, getting around the problem.


shepda (approved):
LGTM

PR #164, opened 2021-04-01
Remove time comparison from ChangeDueDate test

jgar123:
A decision is yet to be made regarding what times are rendered in the UI and what is stored in the database. As such, the test of time equality in ChangeDueDate has been removed where only testing the change in day, month and year is tested. A ticket will be raised to investigate this further.


jennikate:
User can't set time in the due date field, only date.
Currently time in code is using `local time`, ticket being opened to get confirmation on whether we want `local` or `utc` & code will be updated from there

jennikate (approved):
👍

PR #169, opened 2021-05-13
COP-4073 Add styling for disabled inputs

jennikate:
*AC*
Some disabled inputs are not using the correct grey to indicate disabled field -- fix

*Notes*
- updated formio-gds-templates package and published
- updated form override scss and moved it to top level assets as forms within tasks were not receiving the style

*To test*
- Delete `node-module`  & `package-lock`
- run `npm i`
-  run `cop-ui` connected to dev
- view the NS Referral form
- > the businessKey field should be dark grey (background-color: #e9ecef / opacity: 1)
- go to a `Target Information Sheet` task (you may need to create one from Cerberus and assign it to you via Cockpit.)
- > Port, Mode, and other disabled fields should be dark grey


jgar123 (approved):
LGTM

PR #171, opened 2021-05-21
Fix "textarea" div line breaks and Prettier config

gavinorland:
This adds the same SCSS as in V1 to fix the display of line breaks in Case View.

Also a fix to the Prettier config and some corresponding file formatting.


shepda (approved):
LGTM

PR #174, opened 2021-05-24
COP-6752: Update tasks assigned to you page to be same as v1

ajames14:
This branch updates the "Tasks Assigned to you" page to reflect the format of V1. Full details of the discrepancy are found here: https://support.cop.homeoffice.gov.uk/browse/COP-6752

**To Test:**

- pull and run locally pointing proxy to dev
- navigate to "Tasks assigned to you page" via the dashboard. If you don't have any tasks you'll need to create one (complete a Record Border Event form to trigger task)
- You should now see an "Action" button rather than Claim/Unclaim and the page should now look the same as V1
-  Clicking Action will take you to the Task Page where you can complete the task

### DIFF
--- a/client/src/pages/tasks/components/TaskList.test.jsx
+++ b/client/src/pages/tasks/components/TaskList.test.jsx
@@ -5,7 +5,7 @@ import TaskList from './TaskList';
 
 describe('TaskList', () => {
   it('renders without crashing', () => {
-    shallow(<TaskList groupBy="category" />);
+    shallow(<TaskList groupBy="category" taskType="groups" />);
   });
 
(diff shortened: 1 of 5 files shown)


client/src/pages/tasks/components/TaskList.test.jsx
```diff
@@ -20,6 +20,7 @@ describe('TaskList', () => {
           },
         ]}
         groupBy="category"
+        taskType="groups"
```
rfndn, line 23: This could be a constant, other than that all looks good :)

rfndn (approved):

PR #180, opened 2021-06-10
COP-7110: COP v2 showing blank page [local storage issue]

ajames14:
This branch updates the Current Group context to store the currentGroup in local storage, rather that secure local storage. This is to fix a bug where by the encryption/decrytion of the current group object was cause V2 UI to break in certain circumstances. 

**To test:**
- run v2 locally with proxy pointing to dev
1. Submit the Report a Variance form
2. Go to tasks and click "Action" on the newly created one
3. Click the link on the first page of the form: 
"To complete this form, you will need to review the case history (opens in new tab)."
4. Navigate back to first window and complete the form
5. Go back to task list, click on Action on the OTHER task the workflow triggered.
6. Click the link on the first page: "To complete this form, you will need to review the case history (opens in new tab)."
7. New browser page will no longer break and everything should continue to work.

### DIFF
--- a/client/src/utils/CurrentGroupContext.jsx
+++ b/client/src/utils/CurrentGroupContext.jsx
@@ -1,6 +1,5 @@
 import React, { createContext, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-import SecureLocalStorageManager from './SecureLocalStorageManager';
 
 export const CurrentGroupContext = createContext({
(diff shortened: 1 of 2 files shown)


rfndn:
LGTM, but the tests are failing. They are not deterministic, so rerunning them might fix the issue.

rfndn (approved):

PR #184, opened 2021-06-14
COP-7382: Change no team set error message

ajames14:
This branch changes the error message for when a user doesn't have a team set up on keycloak telling them how to reach support link.

To test:
- delete your team on my profile page (or via keycloak)
- log out and in again. You should see the correct error message. You will then have to add your team back in (ask me or dev ops to do this)


rfndn (approved):
LGTM

PR #185, opened 2021-06-15
Handle 'cancel-form' custom event

gavinorland:
User Profile form dispatches this event - when it does we should return the user to the dashboard.


shepda (approved):
LGTM

PR #187, opened 2021-06-16
Use direct link to User Profile form

gavinorland:
We're avoiding using Navi for this link as we want COP to refresh so that a new Keycloak token is fetched for the user and they can more quickly see their details refreshed (as in V1).


shepda (approved):
LGTM

PR #195, opened 2021-07-01
COP-7650: Update Case History section so forms display in order with correct Latest label

ajames14:
**Description**
This branch fixes a bug where "Latest" label is showing for all form instances under the Case History section on the Cases page. It puts the forms in the correct submission date order (from latest to oldest) and adds the Latest label for only the latest instance. It also updates the tests for this change. 

**To test**
- Run branch locally pointing proxy to dev. 
- Navigate to cases page. Search for a case which you know has multiple completed forms (e.g. http://localhost:3000/cases/DEV-20210628-1460). Go to Case History tab and select drop down that contains multiple forms (for this example, "Collect Event at Border" at the bottom).
- Click on the Form links one by one to open them. You will see they are now in the correct order with Latest first and a Latest Label on only the first.

### DIFF
--- a/client/src/pages/cases/components/FormDetails.jsx
+++ b/client/src/pages/cases/components/FormDetails.jsx
@@ -23,6 +23,10 @@ const FormDetails = ({ formReferences, businessKey }) => {
   const [submissionData, setSubmissionData] = useState(null);
   const [selectedFormInstance, setSelectedFormInstance] = useState(null);
 
+  // const sortFormReferenceses = formReferences.sort((a,b) => {
+  //   return new Date(b.submissionDate) - new Date(a.submissionDate)
+  // })
+
   const fetchForm = async (formName, formVersionId) => {
     if (axiosInstance) {
       try {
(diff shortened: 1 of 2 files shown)


client/src/pages/cases/components/FormDetails.jsx
```diff
@@ -76,77 +76,86 @@ const FormDetails = ({ formReferences, businessKey }) => {
...
+                    </div>
+                    <div className="govuk-summary-list__row">
+                      <dt className="govuk-summary-list__key">Submitted on</dt>
+                      <dd className="govuk-summary-list__value">
+                        {dayjs(formInstance.submissionDate).format('DD/MM/YYYY HH:mm')}
+                      </dd>
+                    </div>
+                    <div className="govuk-summary-list__row">
+                      <dt className="govuk-summary-list__key">
+                        {index === 0 ? <span className="govuk-tag">Latest</span> : null}
```
gavinorland, line 114: Could be `{!index && <span className="govuk-tag">Latest</span>}`

gavinorland (dismissed):
Looks fine Abi - just the one suggestion there.

gavinorland (approved):

PR #202, opened 2021-08-16
COP-8225 Form content page for targetTask form

jennikate:
## Requirement

Specifically for the targetTask page for RoRo data for now (future requirements unknown) when a FLO claims and views a targetTask in their task list they should first be taken to a page of non-interactable content about that task. From that page they can then enter the form.

NOTE: in the 'warnings' section the prototype shows `, Monitor only` - this data does not exist yet so that part has been removed from this code.

NOTE: until this code is merged to dev I have not removed the current 'non interactable' form page from the targetTask form so as not to interfere with other testing.

## To test

1. Create task(s) in Cerberus
2. Claim Cerberus Task and submit Target Information Sheet
3. go to COP UI
4. Claim the target task you just created
> - you should be taken to a page similar to https://cop.prototype.cop.homeoffice.gov.uk/blue/sprint-63/target-detail
> - the page should be populated with data from the Cerberus task/target information sheet you submitted
5. Click cancel
> - you should be returned to the Dashboard
6. Click on a different task (not a target task)
> - you should be taken to the relevant task form
7. Go back & click on your target task again
8. Click on 'record outcome' button
> - you should be taken to the target task form (see note above re first page)

## Notes on scenarios that are handled
- only required fields entered on the Target Information Sheet
- ALL fields entered on the TIS
- Multiple passengers entered on the TIS
- Multiple nominals entered on the TIS


jgar123 (approved):
LGTM

PR #209, opened 2021-08-26
COP-8540 Make targets a top level component

jennikate:
## Description

Targets as a concept are similar to tasks, but the UX and data needs are different. Agreed with Ratan to make Targets exist on the same level as Tasks, Cases, Reports and therefore have moved the files to their own folder.

Also tidied up the code and imports.

And moved pagination to a common component as it's used by both Tasks and Targets

----

## To test

- check the `Dashboard` you should have `tasks`, `team tasks`, and `targets`
- click on `targets` you should see a list of targets for your current group, along with due, assigned to, and claim/unclaim button
- visit `tasks`, `team tasks`, and `targets` in turn
- > if there are more than 20 tasks the pagination buttons should be active
- > click through pagination buttons to check they are updating the list on the page


jgar123 (approved):
LGTM

PR #212, opened 2021-09-03
COP-7643 remove production label

Ali-Baker:
https://support.cop.homeoffice.gov.uk/browse/COP-7643

- Remove `production` label and keep `dev`
- Add `.env` to `.gitignore`


jennikate (approved):
LGTM to display only on dev
Checking with @Ali-Baker do we want to show label on Staging and SIT as well

jennikate (dismissed):
LGTM

jennikate (approved):

jennikate (approved):

PR #213, opened 2021-09-07
Add checks to handle non RoRo tasks

jennikate:
E2E testing identified a scenario where non RoRo tasks can be created as a target task, but the targetTask form doesn't handle it.

This PR is to set roroData var to empty object if the mode is anything other than RoRo and to hide the RoRo accordion section.

To test
- create a non RoRo target
- go to that target in cop-ui
- page should load


jgar123 (approved):
LGTM

PR #214, opened 2021-09-07
COP-8640 Pagination for Target List

jennikate:
## Updated

Added pagination for the target list based on the designs we use for Cerberus UI
Tidied up Target List code

## Notes

- This does not include the design of the task list for targets

## To test

- Go to cop-ui Targets (/target)
- if there are more than 20 tasks you should have pagination options [1, [x], next, last >>]
- click on any link, it should update the tasks to show ones for that page
- if you are on a page other than 1 you should have the options [<<first, previous, 1, [x], next, last>>]
- if you are on the last page you should not have next/last as options
- on the right hand side you should see the number of the tasks being shown


jennikate:
Superceded by #216 216

PR #218, opened 2021-09-16
COP-8225 Update threat indicator key name

jennikate:
Discovered the key name for indicators had changed - this PR updates that


jgar123 (approved):
LGTM

PR #245, opened 2021-11-05
COP-9152 Swap checkboxes to small version

jennikate:
Small design update - swap checkboxes to the small version for the filters in target list page

Updated
- checkbox class
- checkbox label padding


jgar123 (approved):
LGTM

PR #250, opened 2021-11-22
COP-9249 Intel Workflow - Product Review Date (US 2 - to 'Change' the PRD 

jgar123:
### Description
This PR adds the "change date" functionality for product review dates. As well as this, it refactors the existing test assertions for the `ProductReviewDate` component to test each arm of logic separately.

### To test
- Pull and run against dev
- Navigate to cases and search for `enhance` case
- *You should only see tasks in the product review date accordion that are active processes and have a review date in the future*
- Submit a `record action` form with the review date specified for the future
- Refresh the page
- Click `Change` for the task just created in the `Product Review Date` accordion
- *You should see a date time input appear with the date time pre-populated with the existing date time*
- Change the date time to something else (still in the future) and click save
- *It should submit successfully and produce an alert for 5 seconds*
- *After the alert goes away, the updated review date should be displayed*


jennikate (approved):
LGTM

PR #257, opened 2021-12-02
COP-9479: Force docker build to use node v14

paulfgraham:
### Description
The docker build is failing because `matrix.node-version` got updated somewhere along the line to use v16, which is incompatible with some of the npm modules we are using in COP UI. Therefore, force it to continue using v14 to avoid the need for full regression testing.

For the sake of clarity and avoidance of doubt, the first build we saw that was using v16 was today - every build prior to today was using v14. So rather than being a backwards step (@michaelruocco's point), we're simply not stepping forward. 🙂 

### To test
If the docker build works in https://github.com/UKHomeOffice/cop-ui/actions, this has succeeded. If not, it has failed.


michaelruocco:
Not a reason to block anything, just just for my own curiosity more than anything, do we have any idea of how much work would be involved in enabling cop ui to work with node v16 or higher.

Appreciate it might be a bit of an unknown in terms of exactly how much work is required, but do we have a regression test suite that would give us some clues and if so, how much coverage does that suite give?

Only reason I ask is it seems like previously the pipeline was able to run using v16 but I guess now we are forcing ourselves to use a version thats two major versions behind that, which seems a bit of a backwards step. Do we know what has changed that had introduced this compatibility or was it always there?

paulfgraham:
@michaelruocco:

Actually, no, the previous pipeline - the one that ran yesterday - ran using v14. The version seems to have gone up between the build yesterday and the one this morning.

In terms of how much work is involved, it's difficult to assess at this point. v16 at the _least_ requires a new major version of `node-sass`. Considering that is for the styling of the whole UI, and considering it is a new major release, it would require a fair bit of regression testing to make sure nothing weird has happened visually. That's even assuming that new version has not deprecated - i.e., removed - anything we're actually relying on with our .scss files.

I don't know if there are any other npm modules that would also need to be upgrade - it was a pretty hefty error log and the only one I saw with a cursory inspection was `node-sass`.

paulfgraham:
Let me quickly clarify further - we didn't introduce anything between the two builds that would change the version of node. It seems to be an environment variable - `{ matrix.node-version }` - that is picked up by the GitHub Action that has changed. What this change does is force it back to what it was _yesterday_, rather than accepting the upgrade.

I don't know whether that upgrade is GitHub-wide or whether someone has deliberately gone in and changed it to be v16... but I don't see any PRs moving it up. 🤷

Ali-Baker (approved):

michaelruocco:
Yeah that makes sense, I just had a look around and it has been a github wide upgrade by the looks of things, these links have some details:
https://github.com/actions/virtual-environments/blob/releases/ubuntu20/20211129/images/linux/Ubuntu2004-README.md
https://github.com/actions/virtual-environments/issues/4446

paulfgraham:
Thanks for finding that, @michaelruocco! That was a proper breaking change from our point of view.

PR #259, opened 2021-12-12
COP-9604 Clear filter not working as expected on COP targets page

charles474:
## Description
This PR fixes a bug where the clear filter button did not reset target count on the first try, required two tries to reset the target count (COP-9604).

* JIRA Ticket: https://support.cop.homeoffice.gov.uk/browse/COP-9604

## To Test
- Pull and run against dev
- On cop-ui, go to targets (**'Targets assigned to your teams'** box)
- Note down the **total targets** displayed
- Apply more than 1 filter and check the number of targets
- Clear all filters
- Check the total targets displayed

## Developer Checklist

\* Required

- [ ] Up-to-date with main branch? \*

- [ ] Does it have tests?

- [ ] Pull request URL linked to JIRA ticket? \*

- [ ] All reviewer comments resolved/answered? \*


Shaks (approved):

client/src/pages/targets/TargetsListPage.jsx
```diff
@@ -231,9 +231,9 @@ const TargetsListPage = () => {
      * reset groups selected to null
      * uncheck any checked checkboxes
      */
+    localStorage.removeItem('checkbox', groupsSelected);
```
michaelruocco, line 234: looks like a pretty simple change, just out of interest was it not possible to add a unit test that recreated the bug before applying the fix?

charles474: The previous test should have picked it up but I believe because the mock fixture always returned the same thing it slips through the cracks. Currently have it on a to-do list to find a better way of replicating that bug.

michaelruocco: Okay no worries, I don't have much to offer by way of help either as I am not that up to speed on react testing either.

I don't think it needs to block this PR if you want to get it merged and fixed. Might be worth checking with @jgar123, @jennikate or @Shaks to see if they have any ideas, but if not we can always try and add another test later if needed.

PR #263, opened 2022-02-03
COP-10006 - Event at border auto populate

charles474:
This PR supplies the `recordBorderEvent` form with the pre-fill data for when a negative outcome has been recorded.

Notes: This approach is not the best however for V2, work will be carried out to refactor the logic of how we are providing pre-fill data to the form.


jgar123 (approved):
LGTM, this change is temporary in order to satisfy a high priority requirement. This code will be removed once the repo has migrated to the new component library and our means of generating prefillable data changes (which will be a feature of work to come)

PR #277, opened 2022-02-11
COP-9749: Display COP React Form Renderer when appropriate

paulfgraham:
### Description
Added dependencies for the COP React Form Renderer and COP React Component libraries, along with changes to the `DisplayForm` to be able to render either a FormBuilder-generate form (with Form.io) or a new COP Form (with the CRFR).

https://support.cop.homeoffice.gov.uk/browse/COP-9749

### To test
The functionality introduced in this PR is presently not testable as there is no way for COP UI to source the new style of form and therefore no way to see the CRFR in play.

However, it's still worth testing that the existing forms continue to display and behave fine.


philjlharvey (approved):

PR #278, opened 2022-02-14
COP-10124: Add separate phase banner

paulfgraham:
### Description
The phase banner was part of the header but it should be its own thing.

https://support.cop.homeoffice.gov.uk/browse/COP-10124

### To test
Covered by accompanying unit test, which was modified from the unit tests for the `<Header />` component.


philjlharvey (approved):

Ali-Baker (approved):

PR #280, opened 2022-02-14
COP-10124: Add Logo component

paulfgraham:
### Description
Added an svg file for the Home Office logo, taken directly from https://design.homeoffice.gov.uk/components?name=Header, and a wrapper component for rendering it wherever necessary (currently this will be just in the `<Header />` component but it could feasibly be displayed elsewhere too (reports, etc.)

https://support.cop.homeoffice.gov.uk/browse/COP-10124

### To test
There's an accompanying unit test to make sure it's actually showing the right svg but the main test will be that it renders appropriately in the header and scales accordingly when the header is resized.


philjlharvey (approved):

Ali-Baker (approved):

PR #281, opened 2022-02-14
COP-10124: Align Header to HO styling

paulfgraham:
### Description
Renamed `index.jsx` to `Header.jsx` (and also the folder goes from `header` to `Header`, which is more appropriate) and added a barrel file instead to make it more consistent with the patterns elsewhere.

The `<Header />` itself is now in the Home Office style with the `<PhaseBanner />` moved to a separate component and the Home Office logo showing in the top-left corner.

### To test
The existing unit tests have been updated but the main test will be to check that it renders appropriately and the contents scale accordingly when the window is resized.


philjlharvey (approved):

PR #288, opened 2022-02-17
COP-10170: Change header link to "Your profile"

paulfgraham:
### Description
Bug reported that the header link still says "My profile" but should say "Your profile". This has now been changed.

### To test
Associated unit tests have been updated. Otherwise, look in the UI to make sure the first link in the header says "Your profile".


philjlharvey (approved):
Reviewed by Ali & Keenan

PR #291, opened 2022-02-21
COP-8191 Force nominal type to be capitalized

jennikate:
## Description

The nominal type is provided all in lowercase. It should be capitalized.

## To test

- open a target on cop-ui that has a nominal
- the nominal type (e.g. account) should be capitalized (e.g. Account)


Shaks (approved):
LGTM

PR #294, opened 2022-02-25
COP-10205: Set max-width to 1160px

paulfgraham:
### Description
Introduced a new class for `hods-width-container` to replace `govuk-width-container` and swapped it out in the various places that used it. This new class has `max-width` set to `1160px`, rather than the `1400px` of the `govuk-width-container`.

### Note
The Figma designs associated with the JIRA ticket say `1220px` but Will from the UCD team has confirmed it should be `1160px`, the same as https://design.homeoffice.gov.uk/.


gardnersam (approved):
Looks good to me

PR #295, opened 2022-02-25
COP-10205: Improve way of overriding link colour

paulfgraham:
### Describe
The old way of changing the link colour involved `!important` in the CSS, which is always a last resort. This is a much cleaner way of overriding the colour.

### To test
All links within COP UI should have the colour `#005ea5`.


client/src/assets/styles/hods.scss
```diff
@@ -1,6 +1,16 @@
...
+$govuk-secondary-text-colour: #505a5f;
 
 @import '~govuk-frontend/govuk/all';
+
 .hods-width-container {
   @include govuk-width-container($govuk-page-width);
 }
+.govuk-link,
+.govuk-link:link {
+  color: #005ea5;
```
gardnersam, line 15: can this use the `$govuk-link-colour` defined above?

paulfgraham: Yes it surely can! 🤦‍♂️

gardnersam (approved):
Looks good to me

PR #296, opened 2022-02-25
COP-10205: Fix phase banner

paulfgraham:
### Description
The tags should be in uppercase and the environment tag shouldn't show in production.


client/src/components/PhaseBanner/PhaseBanner.jsx
```diff
@@ -3,20 +3,24 @@ import { Link, Tag, Utils } from '@ukhomeoffice/cop-react-components';
...
 const DEFAULT_CLASS = 'govuk-phase-banner';
 const PhaseBanner = () => {
   const { t } = useTranslation();
   const uiVersion = config.get('uiVersion');
-  const uiEnvironment = config.get('uiEnvironment');
+  const uiEnvironment = (config.get('uiEnvironment') || '').toLowerCase();
   const serviceDeskUrl = config.get('serviceDeskUrl');
 
   const classes = Utils.classBuilder(DEFAULT_CLASS, []);
+  const showEnvironment = uiEnvironment.indexOf('prod') < 0;
```
gardnersam, line 17: I think this name is confusing, in the context of the PR I understand you want to hide the tags in prod but this takes maybe 2 seconds of reading to figure out otherwise. Could this be called inProd? So you immediately know what it's value means.

client/src/components/PhaseBanner/PhaseBanner.jsx
```diff
@@ -3,20 +3,24 @@ import { Link, Tag, Utils } from '@ukhomeoffice/cop-react-components';
...
 
+// Styles
+import './PhaseBanner.scss';
+
 const DEFAULT_CLASS = 'govuk-phase-banner';
 const PhaseBanner = () => {
   const { t } = useTranslation();
   const uiVersion = config.get('uiVersion');
-  const uiEnvironment = config.get('uiEnvironment');
+  const uiEnvironment = (config.get('uiEnvironment') || '').toLowerCase();
```
gardnersam, line 13: I don't know what comes out of config.get but why not transform it to upperCase here straight away instead of in the scss later?

gardnersam (commented):
Few questions, one for potential readability and another for possibly unnecessary double case swap. Variable name certainly not a blocker.

client/src/components/PhaseBanner/PhaseBanner.jsx
paulfgraham: Ah, it just has to go one way or the other - the transform in the CSS is to override an existing transform to sentence case so it's needed there whether this is `toLowerCase()` or `toUpperCase()`.

client/src/components/PhaseBanner/PhaseBanner.jsx
paulfgraham: Okay, I'll change it. 🙂

gardnersam (approved):
Looks good, thanks for the name update

PR #297, opened 2022-02-25
COP-10205: Fix layout and colouring on Home screen

paulfgraham:
### Description
The padding and margins needed adjusting, the font size wasn't scaling between mobile and tablet/desktop views appropriately, and the cards on the Home screen also didn't scale with their contents - they were all the same size as each other.

Further, the background colour for the `body` was wrong - I had used the background colour for Figma rather than the _design_ on Figma. 🤦‍♂️ 

### To test
The layout should match the Figma designs from the JIRA ticket.


gardnersam (approved):
Looks good

PR #303, opened 2022-03-03
COP-7968: Handle drafts on Task Page

paulfgraham:
### Description
Draft forms have a slightly different structure to tasks - they don't have a `form` property so that now gets loaded separately - and they also don't have a due date so that also needed to be taken account of.

### To test
When clicking on Continue from the list of in progress forms, the form should now correctly load without any exceptions.


client/src/pages/tasks/TaskPage.jsx
```diff
@@ -56,12 +57,21 @@ const TaskPage = ({ taskId }) => {
           // Spread the taskData into seperate variables
           const {
             variables,
-            form,
+            form: formInfo,
```
keenan218, line 60: Why is `form: formInfo` needed and how's it different from `form`

paulfgraham: Lines 68 to 73 are why. If we keep it called `form` at this point, it's a `const`... and it is `undefined` for draft forms. If we don't use a `formInfo` alias, we'll be stuck with an undefined `form` to display in the renderer, which will cause an exception.

The logic from lines 68 to 73 will go and get us the `form` JSON for draft forms, which can then be used by the renderer without an exception.

paulfgraham: Do you think the comment on Line 66 needs additional detail?

keenan218: Ah okay i understand now, and yeah maybe just a brief explanation of form being undefined for draft forms, hence why we need the alias.

keenan218 (approved):

PR #304, opened 2022-03-04
COP-10205: Remove theme switching for Roboto

paulfgraham:
### Description
Roboto is looking good so there's no need to have the theme switcher in place any longer to do the quick comparisons between it and GDS Transport.

### To test
Roboto should show in all circumstances - previously, it would only show if the Dark theme was enabled on the OS or in the browser.


client/src/assets/styles/fonts.scss
```diff
@@ -1,26 +1,24 @@
...
+}
 
-  $govuk-font-family: 'Roboto', arial, sans-serif !important;
+$govuk-font-family: 'Roboto', arial, sans-serif !important;
 
-  body {
-    font-family: 'Roboto', arial, sans-serif !important;
-  }
+body {
+  font-family: 'Roboto', arial, sans-serif !important;
```
p3dr0migue1, line 17: Maybe a stupid question, but if both font faces are called `'Roboto'` then how do you know which one are you calling? The bold one or the normal one? 

Should they be called instead `'Roboto'` and `'Robot-Bold'` to make it easier to understand which is being used?

charles474 (approved):
lgtm

client/src/assets/styles/fonts.scss
paulfgraham: It's based on the font weight - the 400 is regular, 700 is bold. If you name them differently, you have to use the different name when you want a bold font. :)

p3dr0migue1 (approved):

PR #305, opened 2022-03-07
COP-10265: Show underline on hover of title

paulfgraham:
### Description
The title link wasn't showing an underline when being hovered on. This is now fixed.

https://support.cop.homeoffice.gov.uk/browse/COP-10265

### To test
Make sure the link in the title ("Central Operations Platform") now shows an underline when it is being hovered on. When it is not being hovered on, there should be no underline, unlike other links.


client/src/components/Header/Header.scss
```diff
@@ -63,7 +63,7 @@ $govuk-font-family: 'Roboto', arial, sans-serif;
     &__link,
     a {
       display: inline-block;
-      text-decoration: none !important;
+      text-decoration: none;
```
p3dr0migue1, line 66: YES!! Nice one

p3dr0migue1 (approved):

PR #308, opened 2022-03-08
Adjust build and deployment steps

paulfgraham:
### Description
The changes in `green-dev` were very specific to that branch and incompatible with `master`. Adjusting those changes to be able to sit alongside those already in `master`.


azee781 (approved):
seems fine.

PR #309, opened 2022-03-09
COP-10177 target screen

charles474:
This PR adds the new design for the target capture screen as per the prototype.

There is still a fair amount of work to be done but this allows us to test some of the tickets that make up this target screen.

> Note: There are still `TODOs` littered all over the place which will be resolved once work is complete.


client/src/utils/TabSelectedContext.test.jsx
```diff
@@ -0,0 +1,14 @@
...
+import { TaskSelectedTabProvider } from './TabSelectedContext';
+
+describe('TabSelectedContext', () => {
+  it('can render components without crashing', async () => {
+    const wrapper = mount(
+      <TaskSelectedTabProvider>
+        <div>Hello</div>
+      </TaskSelectedTabProvider>
+    );
+    expect(wrapper).toBeDefined();
```
michaelruocco, line 12: might just be my ignorance but what is this actually testing?

client/src/pages/targets/__components__/TargetsListPage/TargetListCardRow.jsx
```diff
@@ -0,0 +1,119 @@
+import React from 'react';
```
michaelruocco, line 1: I am okay if we do it later, but I don't see any tests for this component?

client/src/pages/targets/__tests__/__snapshots__/Tabs.test.jsx.snap
```diff
@@ -0,0 +1,538 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
michaelruocco, line 1: what are these .snap files for? I don't think I've seen them used before?

client/src/pages/targets/utils.js
```diff
@@ -18,8 +18,8 @@ export const formatCategory = (category) => {
   return category.slice(-1);
 };
 
-export const formatDate = (date) => {
-  return dayjs(date).format('DD MM YYYY HH:mm');
+export const formatDate = (date, withFormat = 'DD MM YYYY HH:mm') => {
```
michaelruocco, line 21: should be easy enough to add a test for this one I think?

charles474: the snaps are basically screenshots of the page, similar to when you do `ReactDOMServer.renderToSting(...)`

charles474: just saved having to dump a whole load of HTML in there

client/src/utils/TabSelectedContext.test.jsx
charles474: just testing that the page does get rendered and doesn't throw an error, equally i can always just test for the presence of the `hello` string.

client/src/pages/targets/__tests__/__snapshots__/Tabs.test.jsx.snap
charles474: same response as here
https://github.com/UKHomeOffice/cop-ui/pull/309#discussion_r822448886

client/src/pages/targets/__components__/TargetsListPage/TargetListCardRow.jsx
michaelruocco: if its covered thats probably okay, just looked like a fair bit of functionality in there, but if you are happy its properly tested thats cool.

michaelruocco (approved):
LGTM

PR #313, opened 2022-03-09
COP-10232: Add date and number format utilities

paulfgraham:
### Description
Added utility methods for formatting dates in a variety of ways, along with unit tests for them. One of those formats necessitated ordinal numbers, so also added a utility method (and tests) for that.

https://support.cop.homeoffice.gov.uk/browse/COP-10232

### To test
There's nothing directly to test in this PR for the QA team - the utility functions have accompanying unit tests with 100% coverage.


client/src/utils/Format/formatNumber.js
```diff
@@ -0,0 +1,18 @@
+const PLURAL_RULES = new Intl.PluralRules([], { type: 'ordinal' });
+const SUFFIXES = new Map([
+  ['one', 'st'],
+  ['two', 'nd'],
+  ['few', 'rd'],
```
keenan218, line 5: Given the first two maps should the third one not be three instead of few?
```suggestion
  ['three', 'rd'],
```

keenan218 (commented):
A quick one re the mapping of numbers. For 11 12 and 13, would these not come up as 'st' 'nd' and 'rd', or is there something built into day.js which doesn't require you to specify different behaviour for those numbers and automatically changes them to 'th'?

client/src/utils/Format/formatNumber.js
paulfgraham: I thought the same thing, @keenan218. I changed it at first myself but that's apparently not how `Intl.PluralRules` works. "few" = "three" for some reason. No idea why. 🤷‍♂️

keenan218: Ah okay, if it doesn't like the use of 'three' we shall just stick with 'few', as long as it works the same i guess it doesn't matter too much...

keenan218 (approved):

PR #316, opened 2022-03-10
COP-10177 target screen v1

charles474:
This PR is a duplicate of a previous PR that was merged and reverted due to `green-team` merge but this essentially adds the new design for the target capture screen as per the prototype.

There is still a fair amount of work to be done but this allows us to test some of the tickets that make up this target screen.

Note: There are still TODOs littered all over the place which will be resolved once work is complete.


Shaks (approved):
LGTM

client/src/assets/styles/index.scss
```diff
@@ -53,25 +55,184 @@ div[ref='input'] {
...
+    margin-top: -2em;
+  }
+
+  .highlighted-row {
+    background-color: rgb(255, 236, 199);
+  }
+}
+
+.moj-badge {
+  font-family: 'GDS Transport', arial, sans-serif;
```
paulfgraham, line 165: Errr... this is not the correct font anymore. Also, why are we using the MOJ Badge, rather than the Home Office `<Tag />` from https://github.com/UKHomeOffice/cop-react-components?

client/src/assets/styles/index.scss
```diff
@@ -53,25 +55,184 @@ div[ref='input'] {
...
+    color: #144e81;
+    background: #d2e2f1;
+  }
+
+  &--large {
+    @include govuk-font($size: 16, $weight: 'bold');
+  }
+}
+
+.govuk-link-button {
```
paulfgraham, line 203: This is a pretty global change here and will be application-wide. It will have farther-reaching implications than just the target screen...

client/src/assets/styles/index.scss
```diff
@@ -53,25 +55,184 @@ div[ref='input'] {
...
+  background: none;
+  border: none;
+  padding: 0 !important;
```
paulfgraham, line 206: Is it _that_ `!important`? We should really try to avoid using this wherever possible... that applies to the uses of it below.

client/src/pages/targets/__components__/TargetsListPage/TargetListCard.jsx
```diff
@@ -0,0 +1,88 @@
...
+  checkedRadioBtnName,
+  selectedTask,
+}) => {
+  return (
+    <div>
+      <table className="govuk-table">
+        <caption className="govuk-table__caption govuk-table__caption--m govuk-!-font-size-24 govuk-!-font-weight-regular">
+          {`${targetTaskByVessel.vesselName}, Inbound due ${formatDate(
+            targetTaskByVessel.due,
+            'DD/MM/YYYY HH:mm'
```
paulfgraham, line 28: Why does this particular view get a different date format? It's a UI question more than a dev one... but we should be consistent across the application with these. The only difference between this and the _default_ date format is with the slashes.

client/src/pages/targets/__components__/TargetsListPage/TargetListCard.jsx
```diff
@@ -0,0 +1,88 @@
...
+          )}`}
+        </caption>
+        <thead className="govuk-table__head">
+          <tr className="govuk-table__row">
+            {tableHeaders.map((header) => {
+              return (
+                <th key={uuidv4()} scope="col" className="govuk-table__header">
```
paulfgraham, line 35: You shouldn't use `uuidv4` to generate keys in an array - it satisfies the linter but it will cause needless re-renders. The whole point of the key is to uniquely identify a given item but, while this is unique, it will change every time when it _should_ be stable.

client/src/pages/targets/__components__/TargetsListPage/TargetListCardRow.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+const getStatusLabel = (task) => {
+  if (
+    task.taskStatus === '' ||
+    task.taskStatus === undefined ||
+    task.taskStatus !== TARGET_STATUS_ACKNOWLEDGED
+  ) {
+    return <span className="moj-badge moj-badge--red moj-badge--margin-bottom-2">NEW</span>;
+  }
+  return <span />;
```
paulfgraham, line 20: Why an empty span? Why not just `null`?

client/src/pages/targets/__components__/TargetsListPage/TargetListCardRow.jsx
```diff
@@ -0,0 +1,119 @@
...
+};
+
+const getTargetActions = (task) => {
+  const targeterActions = task?.roro?.details?.targeterActionsRemarks;
+  if (targeterActions.toUpperCase() === 'Monitor only'.toUpperCase()) {
```
paulfgraham, line 25: Magic string? I'm surprised this didn't get picked up by the linter.

client/src/pages/targets/__components__/TargetsListPage/TargetListCard.jsx
charles474: This was more with the UI design we have

client/src/pages/targets/__components__/TargetsListPage/TargetListCardRow.jsx
```diff
@@ -0,0 +1,119 @@
...
+              onChange={(e) => handleRadioButton(e, task, menuName)}
+              checked={isChecked(checkedRadioBtnName, radioBtnIdentifier)}
+              data-testid={radioBtnIdentifier}
+            />
+            <label
+              className="govuk-label govuk-radios__label"
+              htmlFor={`radio-${task.parentBusinessKey.businessKey}`}
+            >
+              &nbsp;
+              <span className="visually-hidden">{task.parentBusinessKey.businessKey}</span>
```
paulfgraham, line 56: If you add this to the top of the file:
`import { VisuallyHidden } from '@ukhomeoffice/cop-react-components';`

You can then change this to:
`<VisuallyHidden>{task.parentBusinessKey.businessKey}</VisuallyHidden>`

That way, we're consistent in how we do it everywhere.

client/src/pages/targets/utils.js
```diff
@@ -18,8 +18,8 @@ export const formatCategory = (category) => {
   return category.slice(-1);
 };
 
-export const formatDate = (date) => {
-  return dayjs(date).format('DD MM YYYY HH:mm');
+export const formatDate = (date, withFormat = 'DD MM YYYY HH:mm') => {
```
paulfgraham, line 21: As I mentioned in an earlier comment, this doesn't make sense to me as we should be consistent with our date formats across the application - we either _always_ have slashes or we never do. 🤷‍♂️

client/src/pages/targets/__components__/TargetsListPage/TargetListCardRow.jsx
```diff
@@ -0,0 +1,119 @@
...
+  if (targeterActions.toUpperCase() === 'Monitor only'.toUpperCase()) {
+    return <span className="moj-badge moj-badge--blue">{targeterActions}</span>;
+  }
+  return <span>{targeterActions}</span>;
+};
+
+const TargetListCardRow = ({ task, handleRadioButton, menuName, checkedRadioBtnName }) => {
+  const radioBtnIdentifier = `target-list-${task.parentBusinessKey.businessKey}`;
+  return (
+    <tbody className="govuk-table__body">
```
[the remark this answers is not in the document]
charles474: I've made a note of this one, if our design changes then this will be on the cards.

PR #319, opened 2022-03-11
Cop 10177 target screen view details

charles474:
# Description
This PR ties up the view details & record outcome actions with the back end service.

# Test
- Pull and run against dev
- Can only access the page locally on this URL [http://localhost:3000/target-capture-screen](url)


client/src/pages/targets/__components__/TargetsListPage/TasksTab.jsx
```diff
@@ -87,38 +93,36 @@ const TasksTab = ({
     setMenuName('');
   };
 
-  /**
-   * TODO Do whatever else that needs to happen here
-   */
+  // TODO Requires implementation
```
michaelruocco, line 96: doesn't need to be in this PR, but I am pretty sure just hitting the acknowledged endpoint will move it back to being acknowledged instead of captured. Or if we want to go back to new, we will likely need another endpoint. Probably best to leave it for now and raise with Peter around which status it should be reverted back to.

michaelruocco (approved):
LGTM

michaelruocco (approved):
LGTM

PR #321, opened 2022-03-14
COP-10264: Fix govuk style overrides (green-dev)

paulfgraham:
### Description
Repeated importing of `.../govuk/all` in the .scss files was overriding the changes being made in `hods.scss`, which kinda defeats the point of it.

Given that `hods.scss` imports `.../govuk/all` itself, the other places can simply import `hods.scss` and should behave themselves just fine while all of the HODS overrides remain in place.

https://support.cop.homeoffice.gov.uk/browse/COP-10264

### To test
Check the secondary buttons on the Case view and nested further into the forms. This was originally noticed on the "People" form within E@B.

### Note
This is identical to https://github.com/UKHomeOffice/cop-ui/pull/318, but this is for `green-dev` only.


client/src/pages/tasks/components/__styles__/TaskListItem.scss
```diff
@@ -1,6 +1,4 @@
-$govuk-page-width: 1400px;
```
gardnersam, line 1: hods.scss sets `$govuk-page-width: 1160px;`, do we definitely want to swap to that here? The PR description sounds like there shouldn't be functional changes

client/src/pages/targets/__components__/Pagination.scss
```diff
@@ -1,10 +1,7 @@
-@import '~govuk-frontend/govuk/all';
-$govuk-link-colour: #005ea5;
-$govuk-link-hover-colour: #003078;
-$govuk-link-visited-colour: #4c2c92;
-$govuk-link-active-colour: #0b0c0c;
 $govuk-secondary-text-colour: #505a5f;
```
gardnersam, line 1: should this be moved to hods.scss?

paulfgraham: Interesting, it's actually exactly the same as the colour that's already being used for that variable in the govuk colour palette, it's known as "dark-grey" and has a hex value of #505a5f. This might be here because there's also a legacy colour and maybe that was causing peculiar behaviour in this view.

However, the redesign to align it with Home Office styling didn't specify anything for this colour so it's not one of the changes we've made... but I've left it here so that it only applies to the Pagination, in case of that "peculiar behaviour" thing. 🤷‍♂️

gardnersam (approved):
Looks good to me, happy with answers to the comments

PR #322, opened 2022-03-14
COP-10177 target screen group counts

charles474:
# Description

This PR reduced the number of calls made to obtain the group counts to a single call.

# Test
- Pull & run against dev
- Load up the old targets screen in one tab, and load up the new target screen in another tab.
- In the network activity, observe the number of calls made between both tabs opened.


jennikate (approved):
LGTM

PR #323, opened 2022-03-14
fixes a situation where localStorage filter value can return as an array of empty string

charles474:
This PR fixes a scenario where localStorage can return an array with an empty string for mode filters.


michaelruocco (approved):
LGTM

PR #324, opened 2022-03-14
COP-10177 target screen sorting

charles474:
This PR adds the sorting params to the task fetch request.


robBrownGitHub (approved):

michaelruocco:
can add some later but I am guessing we dont have any tests asserting that the API calls are being issued correctly otherwise they would need updating here wouldn't they?

michaelruocco (approved):
LGTM

PR #325, opened 2022-03-15
COP-10177 - updates the category sort order from desc to asc

charles474:
This PR reverses the category request param value from `desc` to `asc`


Shaks (approved):

michaelruocco (commented):
LGTM

PR #326, opened 2022-03-15
COP-10177 - Adds support for enabling or disabling new target screen via env variable

charles474:
# Description
This PR adds the capability of enabling the new targets screen so when users access this URL [https://www.dev.cop.homeoffice.gov.uk/targets](url), depending on what the environment flag says, they will either be presented the new target screen or the old target screen.

The `env` variable is `REACT_APP_NEW_TARGET_SCREEN`. This can either be `true` | `false` or not be supplied at all which will equate to `false`.

New environment variable
```
REACT_APP_NEW_TARGET_SCREEN
Options: true | false | can be omitted entirely
```

### Supplemental

This PR also fixes a bug where by the `next` link would be rendered on pagination component when there are no results.

This PR also adds caching of the tabs when a user navigates away to the view details page or record outcome and then back backtracks to the target capture screen.

# To Test
**Step 1:** Pull

**Step 2:** Set up the required configurations (locally)

**Step 3:** In the `.env` file in your local environment, add the following `REACT_APP_NEW_TARGET_SCREEN=true` and start the ui locally. 

**Step 4:** Load up [http://localhost:3000](url), locate the box that contains text `...targets assigned to your team`, click on this. if you had the `REACT_APP_NEW_TARGET_SCREEN=true`, you will be presented with the new targets list page.

**Step 5:** If you had `REACT_APP_NEW_TARGET_SCREEN=false` or `REACT_APP_NEW_TARGET_SCREEN=` or not supplied it at all, you would be presented with the old targets list page (testing this scenario would require you to stop the server, make the needed changes and spin it back up again and repeat **Step 4** observe the behaviour of the page).

> Note: The new targets list page can also be accessed via [http://localhost:3000/target-capture-screen](url)


Shaks (dismissed):

Shaks:
lgtm

adam-thorpe (approved):

PR #330, opened 2022-03-16
COP-10232: Add DateTime component

paulfgraham:
### Description
Added a `DateTime` component to display on the `DraftFormsList`, which makes the "Created" field on the table accessible.

### To test
There are accompanying unit tests but the true test for this will be to make sure screen readers behave themselves when they encounter to the field, and that they read the date and time out in a sensible way.


client/src/pages/forms/components/DateTime.jsx
```diff
@@ -0,0 +1,20 @@
...
+const DateTime = ({ date }) => {
+  return (
+    <time dateTime={Format.date.iso(date)} aria-label={Format.date.aria(date)}>
+      {Format.date.pretty(date)}
+    </time>
+  );
+};
+
+DateTime.propTypes = {
+  date: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired,
```
keenan218, line 17: I think i have seen before as part of the linting rules that `PropTypes.object` is discouraged because they are very generic. Or perhaps this doesn't apply to cop-ui and another repo instead...

paulfgraham: Ah, good shout. It doesn't get flagged up by the linter in COP UI and there's no `PropType` for a date. I can use `PropType.instanceOf(Date)`, though. Hopefully that'll work with that `onOfType`...

paulfgraham: Fixed.

keenan218 (approved):

keenan218 (approved):

PR #333, opened 2022-03-17
COP-10178: Undo capture link is shown for captured task

Shaks:
Undo capture link would be visible to the user to allow undo captured task

image


charles474 (approved):
LGTM

PR #336, opened 2022-03-17
COP-10366: Fix styling of Targets List page

paulfgraham:
### Description
Moved the targets-specific styles to a new stylesheet, rather than them sitting in the top level `index.scss`.

Also fixed an issue with the _existing_ `TargetsListPage` component, which was still exceeding its bounds and looked odd.


charles474 (approved):
LGTM

PR #337, opened 2022-03-18
COP-10232: Change some content for in-progress forms

paulfgraham:
### Description
Aligned the content with the latest Figma designs and updated the unit tests to account for the changes.


client/public/locales/en-GB/translation.json
```diff
@@ -110,10 +110,8 @@
         "continue": "Continue"
       },
       "none": {
-        "heading": "You have no in progress forms",
-        "will-appear": "In progress forms will appear here when you:",
-        "start-new": "start a new form",
-        "exit": "save and exit that form",
+        "heading": "You have no in-progress forms",
+        "will-appear": "Forms in progress will only appear here once<br/> you have started and saved a new one.",
```
keenan218, line 114: Does this "in progress" need a hyphen like the other three have?

paulfgraham: Ha! No. I had that confirmed by Robin as I asked him the same question directly. In this instance, it's essentially the same as "Forms that are in progress...", which shouldn't be hyphenated.

keenan218 (approved):

PR #341, opened 2022-03-18
Ci 88 v6

robBrownGitHub:


client/src/pages/cases/components/ProductReviewDate.jsx
```diff
@@ -152,14 +152,19 @@ const ProductReviewDate = ({ businessKey, processInstances }) => {
...
       }
     }
+  };
+
+  useEffect(() => {
+    checkProductReviewDate();
   }, [reviewDatetime]);
 
   const handleReviewDateChange = (e) => {
     setReviewDatetime({ ...reviewDatetime, [e.target.name]: e.target.value });
```
gardnersam, line 162: I think there is a risk here of if the user hammers the keyboard too fast you will get unreliable updates. To quote a review comment Paul gave me a few weeks ago: "Basically, there's an asynchronous nature to this and if you fire handleReviewDateChange multiple times in rapid succession (such as someone typing quickly) with your code, you're not guaranteed to capture all the changes. This has caught me out countless times!".

Some more info here: https://reactjs.org/docs/hooks-reference.html#functional-updates

This code may be better:

`setReviewDatetime(reviewDatetime => ({... reviewDatetime, [e.target.name] : e.target.value}));`

client/src/pages/cases/components/ProductReviewDate.jsx
```diff
@@ -152,14 +152,19 @@ const ProductReviewDate = ({ businessKey, processInstances }) => {
...
   };
 
   const handleReviewDateSubmit = async (e) => {
-    if (isDateFormatValid && !isDateAfter18Months && !isDateInThePast) {
+    checkProductReviewDate();
```
gardnersam, line 166: I don't understand why we call this here - the review date time can't have changed can it? Or are we covering if the user somehow types then hits submit before the state updates?

client/src/pages/cases/components/ProductReviewDate.jsx
```diff
@@ -152,14 +152,19 @@ const ProductReviewDate = ({ businessKey, processInstances }) => {
...
+    if (isDateFormatValid === true && isDateAfter18Months === false && isDateInThePast === false) {
```
gardnersam, line 167: Could this be ` if (isDateFormatValid&& !isDateAfter18Months && !isDateInThePast) {`

Doesn't really matter either way, just a suggestion

gardnersam (commented):
Few thoughts for you

gardnersam (approved):
Discussed on slack, looks good to me

PR #344, opened 2022-03-21
COP-10377: Reverse sorting order for draft list

paulfgraham:
### Description
The order should be descending date order - i.e., the most recent should be first.

https://support.cop.homeoffice.gov.uk/browse/COP-10377

### To test
Add some draft forms and make sure that the most recently created one appears at the top, rather than the oldest one at the top.


charles474 (approved):
LGTM

PR #347, opened 2022-03-22
COP-10177 target screen group counts

charles474:
# Description
This PR swaps out a slow endpoint for a new endpoint for getting individual group counts.

### Supplemental 
This PR also resolves an issue where clicking the apply button would issue a duplicate network call.


Shaks (approved):
LGTM

PR #355, opened 2022-04-12
fixes setting date value in test which values when date is january be…

michaelruocco:
…cause months are 0 indexed and 0 is not valid input.

I was tripping over a failing test when running the tests against current master branch, and it was blocking me from pushing any new changes, so this PR is my attempt at correcting the existing issue.


client/src/pages/cases/components/ProductReviewDate.test.jsx
```diff
@@ -360,7 +360,7 @@ describe('ProductReviewDate', () => {
     fireEvent.change(dayInput, { target: { name: 'day', value: `${dayValue}` } });
 
     const monthInput = screen.getByRole('textbox', { name: 'Month' });
-    fireEvent.change(monthInput, { target: { name: 'month', value: `${monthValue}` } });
+    fireEvent.change(monthInput, { target: { name: 'month', value: `${monthValue + 1}` } });
```
michaelruocco, line 363: when days js returns the value from `datePlus19Months.get('month')` the month value is 0 indexed, meaning January is returned as 0, which was causing the tests to fail because 0 is invalid input for the month. This wasn't affecting this test directly, it was actually `ProductReviewDate › should display an error validation message if date is in the past` that was failing but I applied this change to all places where month value is set to ensure it doesn't trip us up again.

client/src/pages/cases/components/ProductReviewDate.test.jsx
```diff
@@ -393,18 +393,18 @@ describe('ProductReviewDate', () => {
     fireEvent.click(changeLink);
 
     const todaysDate = dayjs();
-    const datePlus19Months = todaysDate.subtract(3, 'month');
-    const dayValue = datePlus19Months.get('date');
-    const monthValue = datePlus19Months.get('month');
-    const yearValue = datePlus19Months.get('year');
-    const hourValue = datePlus19Months.get('hour');
-    const minuteValue = datePlus19Months.get('minute');
+    const todayMinus3Months = todaysDate.subtract(3, 'month');
```
michaelruocco, line 396: variable name here was clearly a copy and paste that wasn't updated, so I updated it.

client/src/pages/cases/components/ProductReviewDate.test.jsx
```diff
@@ -435,10 +435,7 @@ describe('ProductReviewDate', () => {
     fireEvent.click(changeLink);
 
     const todaysDate = dayjs();
-    // TODO: Make this ACTUALLY work for yesterday.
-    // It looks like it might be BST vs UTC as it was previously working.
-    // const yesterday = todaysDate.subtract(1, 'day');
-    const yesterday = todaysDate.subtract(2, 'day');
```
michaelruocco, line 441: I think this was an attempt at fixing the issue that was tripping me up here, but didn't actually solve the issue in all cases, I believe the fix I have applied to cater for the month index being zero indexed (i.e. 0-11 rather than 1-12) will fix it more permanently.

client/src/pages/cases/utils/validate/FormDateValidation.js
```diff
@@ -80,7 +80,7 @@ export { isDateMoreThanXMonthsAgo };
  * @returns true if date is before current date and time, even by 1 minute
  */
 function isDateBeforeToday(dateToCheck) {
-  const today = dayjs().subtract(2, 'second').format('YYYY-MM-DD HH:mm');
```
michaelruocco, line 83: I have no idea why we were subtracting two seconds from todays date (and then formatting it without seconds anyway) but it seemed odd and incorrect to me, so I have taken it out.

client/src/pages/cases/components/ProductReviewDate.test.jsx
```diff
@@ -449,7 +446,7 @@ describe('ProductReviewDate', () => {
     fireEvent.change(dayInput, { target: { name: 'day', value: `${dayValue}` } });
 
     const monthInput = screen.getByRole('textbox', { name: 'Month' });
-    fireEvent.change(monthInput, { target: { name: 'month', value: `${monthValue}` } });
+    fireEvent.change(monthInput, { target: { name: 'month', value: `${monthValue + 1}` } });
```
michaelruocco, line 449: same change as described in: https://github.com/UKHomeOffice/cop-ui/pull/355/files#r848411976

client/src/pages/cases/components/ProductReviewDate.test.jsx
```diff
@@ -610,7 +607,7 @@ describe('ProductReviewDate', () => {
...
```
michaelruocco, line 610: same change as described in: https://github.com/UKHomeOffice/cop-ui/pull/355/files#r848411976

charles474 (approved):
LGTM

Thanks for leaving the comments at each change, made it easier to understand what was going on.

charles474 (approved):
LGTM

Thanks for leaving the comments at each change, made it easier to understand what was going on.

PR #358, opened 2022-04-14
Solved bug causing Accordion info/subheading not to render correctly

keenan218:
Small fix to remove an erroneous letter when declaring the subheading for the attachments.


gardnersam (approved):
Looks good

PR #359, opened 2022-04-25
Updated Accessibility Statement as per requirements

keenan218:
The Accessibility statement has had some updates. Changes are defined in the following ticket [COP-10270](https://support.cop.homeoffice.gov.uk/browse/COP-10270)


gardnersam (approved):
Looks good to me

PR #362, opened 2022-04-27
Added in the Product Review Date to act the same as it does currently…

keenan218:
… in dev and higher. Added test and test data.

Product Review Date has been re-added to cop-ui, will be found in the accordion above 'actions' on cases that require a future review. 
Example of case that shows Product review date: DEV-20220426-937


client/src/pages/cases/components/CaseAccordionGroup.jsx
```diff
@@ -5,13 +5,29 @@ import { AccordionItem } from '@ukhomeoffice/cop-react-components/dist/Accordion
...
 
-const CaseAccordionGroup = ({ businessKey, caseActions, getCaseDetails }) => {
+const CaseAccordionGroup = ({ businessKey, caseActions, getCaseDetails, processInstances }) => {
   const { t } = useTranslation();
 
   return (
     <div className="govuk-grid-column-full ">
       <Accordion id="accord1">
+        {caseActions.some(
+          (caseAction) => caseAction?.process['process-definition']?.key === 'recordIntelActions'
```
gardnersam, line 17: caseAction will always be defined I think? Otherwise some() wouldn't run the function.

client/src/pages/cases/components/CaseAccordionGroup.jsx
```diff
@@ -5,13 +5,29 @@ import { AccordionItem } from '@ukhomeoffice/cop-react-components/dist/Accordion
...
+        ) && (
+          <AccordionItem
```
gardnersam, line 19: I don't understand the link between having the 'recordIntelActions' key defined and showing the product review date accordion item

client/src/pages/cases/components/CaseAccordionGroup.jsx
```diff
@@ -5,13 +5,29 @@ import { AccordionItem } from '@ukhomeoffice/cop-react-components/dist/Accordion
...
+            heading={t(
+              'pages.cases.details-panel.case-accordion-group.product-review-date.heading'
+            )}
+            summary={t(
+              'pages.cases.details-panel.case-accordion-group.product-review-date.subheading'
+            )}
+          >
+            <ProductReviewDate businessKey={businessKey} processInstances={processInstances} />
+            <p style={{ 'font-size': '1%', visibility: 'hidden' }}>.</p>
```
gardnersam, line 28: What is the purpose of this tiny invisible full stop?

client/src/pages/cases/components/ProductReviewDate.jsx
```diff
@@ -200,195 +192,177 @@ const ProductReviewDate = ({ businessKey, processInstances }) => {
...
-        id={`productReviewDate-${businessKey}`}
-        className="govuk-accordion"
-        data-module="govuk-accordion"
-      >
-        <div className="govuk-accordion__section">
-          <div className="govuk-accordion__section-header">
-            <h4 className="govuk-accordion__section-heading">
-              <span
-                className="govuk-accordion__section-button"
-                id={`heading-productReviewDate-${businessKey}`}
```
gardnersam, line 213: This bit is gone now, just wanted to check that's on purpose

gardnersam (commented):
Few bits I don't quite understand so looking to clarify

client/src/pages/cases/components/CaseAccordionGroup.jsx
keenan218: This was already here when the functionality was in place previously, it has just been relocated from CaseDetailsPanel.jsx so it works within the accordion.

client/src/pages/cases/components/CaseAccordionGroup.jsx
keenan218: We are only wanting to show the ProductReviewDate accordion when there is an action of RecordIntelActions so it is not showing up unnecessarily. It has the same functionality as the PRD currently does where it only shows when certain conditions are met

client/src/pages/cases/components/ProductReviewDate.jsx
keenan218: Yeah, this is how it was before. It was in it's own accordion. Now it is in the react component accordion and is in the accordion group along with actions, attachments and eventually notes

client/src/pages/cases/components/CaseAccordionGroup.jsx
keenan218: the component doesn’t play too nice with nested react components and it won’t render the full size of the accordion when expanded so it wants a bit of html content at the bottom of the accordionItem that isn’t in another react component

gardnersam (approved):
Thanks for the explanations, looks good!

PR #363, opened 2022-05-03
Cop 10347/claim task

keenan218:
Added the frontend changes to ticket [COP-10347](https://support.cop.homeoffice.gov.uk/browse/COP-10347). Allows the user to claim a task from the Timeline in caseview.

**To Test:**
When a case has uncompleted tasks the user should be able to claim one of these unclaimed tasks from the Case Page. If the task is uncompleted but assigned to another user it should display a warning message advising the user this task is unclaimable.


philjlharvey (approved):
approved after some refactoring to put the not started task detail in a separate component

PR #364, opened 2022-05-06
Feature/cop 10589/full width caseview

keenan218:
Providing increased readability to caseview following the most recent figma design. 

- Grey lines have been added between react form fields.
- Full page width has been utilised for both formIO and React forms, getting rid of wasted space on the left of the screen.
- Changed `Not Started` tag to Dark Grey with White text


aliktb (approved):
Approving after discussing changes offline

PR #372, opened 2022-05-23
CI-105 Fixed uploaded file times displaying in UTC during BST.

LewisClark-HO:
The displayed times are now converted to either BST or UTC using Day.js depending upon the time of year.


client/src/pages/cases/components/CaseAttachments.jsx
```diff
@@ -124,7 +130,12 @@ const CaseAttachments = ({ businessKey }) => {
                         </a>
                       </th>
                       <td className="govuk-table__cell">{submittedEmail}</td>
-                      <td className="govuk-table__cell">{submittedDateTime}</td>
+                      <td className="govuk-table__cell">
+                        {dayjs
```
paulfgraham, line 134: Could you do this in a util function, rather than inline? I strongly suspect this won't be the only place that needs formatting to local time, rather than UTC... and there may well already be other places that do something very similar in the codebase (if not identical).

paulfgraham (changes requested):
I think the formatting should be in a util function rather than inline.

client/src/pages/cases/components/CaseAttachments.jsx
LewisClark-HO: I've moved it to a util function as suggested - thank you!

paulfgraham (approved):
Very nice. Thank you for making the suggested changes, @LewisClark-HO. 🙂

PR #393, opened 2022-06-14
Cop 10246 - enable display of COP react form

fisherrj2:
https://support.cop.homeoffice.gov.uk/browse/COP-10246

In COP UI, there is a DisplayForm.jsx component*, which is the primary place where forms are displayed. In that component, there is a mechanism for deciding whether to use the Form.io renderer or the new COP React Form Renderer, depending on the name of the form.

There are, however, a few other places where this switching logic will also need to be implemented (and the new COP React Form Renderer component should be added to the markup, accordingly):

/client/src/pages/cases/components/CaseAction.jsx


keenan218 (approved):
Approved following an offline discussion

PR #394, opened 2022-06-14
Refactors v1 target screen files

charles474:
This PR moves most v1 target capture screen files into one location, leaving behind anything that is reusable and not tightly coupled to v1 capture screen ahead of airpax development)


Shaks (approved):
lgtm

PR #406, opened 2022-06-19
COP-10822 - Search input

charles474:
This PR adds the search input to the Targets List Page.

> ## Note
The drone build is failing due to statement coverage percentage not meeting the required 90% however, this PR does not concern itself with that. Additional tests will be added in subsequent PRs to follow.


client/src/pages/targets/__styles__/TargetsListPage.scss
```diff
@@ -101,6 +101,7 @@
     grid-template-columns: 1.5fr 1fr auto;
     grid-column-gap: 10px;
     align-items: center;
+    float: right;
```
Shaks, line 104: add a utility class in html for floating left or right rather adding here as a rule. If there isn't such a class in COP-UI repo then create one

Shaks (approved):
lgtm

PR #409, opened 2022-06-21
COP-10822 - Adds tests for the targets list page

charles474:
This PR adds tests for the targets list page (including tests for the components which make up the targets list page).

> **Note**
- Test build should now be passing as the requires statement coverage of 90% has been met.


Shaks (approved):
lgtm

PR #412, opened 2022-06-21
EAB-625 Remove form title from cop screens

LewisClark-HO:
UCD advised that the form title should not be displayed on Form Renderer pages. This is because it conflicted with the H1 titles on those pages.

This change adds a check for the renderer being used and will hide the form title if it's a React Form Renderer page.


aliktb (commented):
I don't believe we need this. The form title can be removed from each individual form instead

aliktb:
> I don't believe we need this. The form title can be removed from each individual form instead

After speaking with team, it is needed. Apologies

aliktb (approved):
Looks good

PR #413, opened 2022-06-22
Cop 10246 render cop form

keenan218:
In Caseview a react Case Action will now be visible to the user, they will be able to step through the react form and upon submission it will have similar functionality to formIO Actions where the page is re-rendered so the newly submitted action is visible in timeline.

To test:
Complete 'COP - first form', enter the Case page and you will see 'COP - Fifth Form' in the Actions section. Submit Fifth form and you will see this completed form at the top of Timeline.


aliktb (approved):
Looks good

PR #419, opened 2022-06-24
removed await before navigation

keenan218:
Removed await to solve error when Case Action submit button is pressed


georgepemberton1998 (approved):
Looks good to me

PR #420, opened 2022-06-27
cop-10246: onAction submit to give correct rerouting

keenan218:
changed rerouting after submitting a caseAction to correctly reroute to the current case page and see the newest form displayed in timeline.


aliktb (approved):
Approved

PR #429, opened 2022-07-07
Feature/cop 9749/ManageMyProfile-view

keenan218:
Adding the ability to view your profile through a react form, This pr is only for viewing profile details, Update functionality will be included in a later ticket.

Further detail outlined in the ticket.
https://support.cop.homeoffice.gov.uk/browse/COP-9749

**To Test** 
Check that you are able to view all fields in 'view profile', the "Are You A Civil Servant?" question will not display an answer as this was not previously a question in the old manage profile form, will be added along with the update ticket


client/src/components/form/DisplayForm.jsx
```diff
@@ -140,6 +142,7 @@ const DisplayForm = ({
       },
       shiftDetailsContext: {
         email,
+        location: currentGroup,
```
aliktb, line 145: Just wondering, should this be `currentGroup`?

client/src/utils/LocationContext.test.js
```diff
@@ -0,0 +1,45 @@
...
+    );
+    expect(screen.getByText(/^Custom/)).toHaveTextContent('null function () { [native code] }');
+  });
+
+  it('can have its value updated', () => {
+    const Consumer = () => {
+      const { location, setLocation } = useContext(LocationContext);
+      useEffect(() => {
+        setLocation({
+          code: 'APOLLO',
```
aliktb, line 33: Out of curiosity, is there any reason you are using `code`?

aliktb (commented):
Looks good but just a couple of comments. Also not sure if you need to commit package-lock given your package.json hasn't been modified

client/src/components/form/DisplayForm.jsx
keenan218: Id meant to remove that, it's not needed, good spot

client/src/utils/LocationContext.test.js
keenan218: i had meant to change this to id, i changed it in hooks test but must have forgot to change it here too

aliktb (approved):
Approved

PR #439, opened 2022-07-18
reverted CI-112 attachment permissions fix in ui

aliktb:
Reverting ui change https://github.com/UKHomeOffice/cop-ui/pull/369 as part of the revised attachment permissions implementation

https://support.cop.homeoffice.gov.uk/browse/COP-10988


keenan218 (approved):
looks good

PR #444, opened 2022-07-26
EAB-633 Add ability to launch POS from hook passed to form-renderer

gardnersam:
Extends the `onSubmit` hook passed to form-renderer to handle a new submit type 'submitToEpms'. This can be initiated with an action on a page formatted like this:

```
"actions": [
     { "type": "submitToEpms", "label": "Submit to ePMS" }
 ],
```

This action type will cause COP UI to take the current payload and use it to initiate the POS BPMN process. It does not prevent or alter the current `onSubmit` functionality from then taking place after.


LewisClark-HO (approved):
Looks good!

PR #446, opened 2022-07-27
Cop 10943 - task list in case view

fisherrj2:
Changes to display task list in case view in CYA style


aliktb (commented):
Looks good, but I believe you will need to update the dependency version of `cop-react-form-renderer` so I will wait for that to be deployed before approving

aliktb (approved):
Looks good

PR #448, opened 2022-07-29
reformat headings and list font sizes following comments from testing

bowerscodes-COP:
reformat headings and list font sizes following comments from test


aliktb (approved):
approving after discussing offline

PR #451, opened 2022-08-03
EAB-695 Add go to EAB1 submit type

gardnersam:
Add new type in the onSubmit hook to handle checking the current form data and conditionally sending the user to EAB1. This can be added to a form with an action like 

`"actions": [
          { "type": "transitionToEab1", "label": "Go to EAB1" }
 ],`

if the data matches certain criteria the user is sent to the start of the EAB1 form, and the current Camunda task is submitted so the process can terminate.


LewisClark-HO (approved):
LGTM

PR #461, opened 2022-08-15
save & return later now renders its own success Panel confirmation page

bowerscodes-COP:
### Description
Clicking on 'Save and Return Later' now renders its own confirmation page, similar to 'Submit'.
To match Figma design: https://www.figma.com/file/HvHakMNbQ9Z17Arw8flhuy/COP-Green-team-project-files?node-id=8693%3A58336

[Link to Jira story](https://support.cop.homeoffice.gov.uk/browse/COP-11610)


client/src/pages/forms/__snapshots__/FormPageSaved.test.jsx.snap
```diff
@@ -0,0 +1,52 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
aliktb, line 1: Not sure what this file is

keenan218: As discussed offline with James, it is a good way to test a page with a lot of static text with no real 'functions'.

keenan218 (approved):
Looks good, works as intended and follows figma implementation

PR #463, opened 2022-08-16
Cop 11471/back link

fisherrj2:
Changes across cop-ui to enable a back link that performs the same as the browser back button, working within the FormRenderer pages.  This works by pushing entries onto the browser history stack using pushState, and listening to window.location.pathname changes via a useEffect hook that sets a newPageId prop on the form renderer.  Link to epic: https://support.cop.homeoffice.gov.uk/browse/COP-11586


client/src/components/layout/index.jsx
```diff
@@ -85,15 +85,15 @@ const Layout = ({ children }) => {
...
                     onClick={async (e) => {
                       localStorage.removeItem(cachedTabKey);
                       e.preventDefault();
-                      await navigation.navigate('/');
+                      await navigation.goBack();
                     }}
                     className="govuk-back-link"
                   >
-                    {t('back')}
+                    {t('pages.forms.back-link')}
```
aliktb, line 96: if this is the back button across the ui, I think it would be better to modify the translation.json 'back' property as opposed to the form page-specific back button. If that makes sense

client/src/components/PhaseBanner/PhaseBanner.jsx
```diff
@@ -12,10 +13,23 @@ const PhaseBanner = () => {
   const uiVersion = config.get('uiVersion');
   const uiEnvironment = (config.get('uiEnvironment') || '').toLowerCase();
   const serviceDeskUrl = config.get('serviceDeskUrl');
+  const navigation = useNavigation();
 
   const classes = Utils.classBuilder(DEFAULT_CLASS, []);
   const inProd = uiEnvironment.indexOf('prod') === 0;
 
+  window.onpopstate = () => {
```
aliktb, line 21: just curious why we need to modify the PhaseBanner?

client/public/locales/en-GB/translation.json
```diff
@@ -191,7 +190,7 @@
         "search-placeholder": "Search form name",
         "search-hint": "Must be greater than 3 characters"
       },
-      "back-link": "Back to forms"
+      "back-link": "Back"
```
aliktb, line 193: see comment on `components/layout/index.jsx`

aliktb (commented):
Just a few comments

client/src/components/layout/index.jsx
fisherrj2: Changed to use the back property

client/src/components/PhaseBanner/PhaseBanner.jsx
fisherrj2: Having tried with window.onpopstate handler in various places what works is placing it in PhaseBanner and having an empty handler in DisplayForm.  Take away either or modify either and we don't get the desired functionality.

aliktb (approved):
approved following comments made

PR #467, opened 2022-08-18
Cop-11705/url-updates

keenan218:
Updating the url for feedback link and IT now link. Using new AWS secret named feedback url.

links to following ticket: [COP-11705](https://support.cop.homeoffice.gov.uk/browse/COP-11705)


aliktb (approved):
Looks good

PR #472, opened 2022-08-22
COP-11409 - Target warnings

charles474:
# Description
This PR addresses a previous incorrect behaviour of displaying a warning banner when there was no warning and rendering a target category within it.

Previously, a target category would be rendered (see below) within the banner however, now, if there is no actual warnings, the banner does not render and if there are warnings, the banner will be rendered along with the warnings and the warnings details if any is present.
image

## Post correction
**RoRo**
Target Category has now been repositioned.
image

**AirPax**
image


client/src/pages/targets/airpax/TargetInfoPage.jsx
```diff
@@ -117,11 +118,13 @@ const TargetInfoPage = ({ targetId }) => {
...
-  const formattedWarnings = warnings ? formatWarnings(warnings) : null;
-
-  const notificationData = [formattedCategory, formattedWarnings].filter(Boolean).join(', ');
-  // Get all persons and then assign uniqueIds to each person
+  const selector = RisksUtil.selectors(RisksUtil.risks(targetData?.informationSheet));
+  const warnings = RisksUtil.warning.get(selector);
+  const formattedWarnings = RisksUtil.warning.hasWarning(RisksUtil.warning.status(warnings))
+    ? RisksUtil.warning.format(RisksUtil.warning.types(warnings))
+    : null;
+  const warningDetails = RisksUtil.warning.details(warnings);
```
FrancisBry, line 126: I think this could be simplified. Instead of calling the RisksUtils 10 times in this file, could there not be just three calls like
```
const hasWarning = RisksUtil.hasWarning(targetData?.informationSheet);
const formattedWarnings = RisksUtil.getFormattedWarnings(targetData?.informationSheet); 
const warningDetails = RisksUtil.getWarningDetails(targetData?.informationSheet);
```

FrancisBry (changes requested):

FrancisBry (approved):

PR #474, opened 2022-08-22
changed feedback link uri params

aliktb:
Changing uri parameters based on updated ticket

https://support.cop.homeoffice.gov.uk/browse/COP-11802


bowerscodes-COP (approved):
LGTM

PR #496, opened 2022-09-05
EAB-887 Fix issue where if no camunda process had started when we sho…

gardnersam:
…uldn't try and end it

When we transition to EAB1 on the second screen of EAB2 (what did you find) the user hasn't clicked 'Save and continue' yet, so no business key or camunda process exists. The existing implementation assumes there is and then throws an error, this checks before to avoid trying to submit a non existent process. 

It also adds an early return statement to the section of onSubmit that deals with the transition, otherwise after we move the user to EAB1 the code here will continue, submit the data and create a camunda process behind our backs.


LewisClark-HO (approved):
LGTM

Deepak-Bagrawat (approved):

PR #500, opened 2022-09-06
COP-10065 - Inbound/Outbound filter

charles474:
# Description
Adds a basic react filter component for the inbound & outbound option to the targets list page.

> Note, this filter currently has no functionality.

### Supplemental
This PR does contain some code refactorings hence the size.

# To Test
- Pull and run against dev
- Load up the following URL http://localhost:3000/airpax/targets
- The Filter component should be rendered with the `Direction of travel` filter with the following options: **`Inbound`**, **`Outbound`**, **`Unknown`**.


Shaks (approved):
lgtm

PR #502, opened 2022-09-07
EAB-962 Add route to EAB1 that catches everyone after a set screen

gardnersam:
For MVP we need to take everyone to EAB1 after they enter item details. I've picked an arbitrary screen here that they will have to have passed after the first transition opportunity but before this one to base the move on. 

We need to do it based on something in form data as the action type and the form data is all we have here. We could theoretically create a whole extra action type that always moves you to EAB1 but I think that would just overcomplicate things.


LewisClark-HO (approved):
LGTM

PR #516, opened 2022-09-12
Allow communication with COP Integration Service and COP Targeting API. Pipeline changes and K8s fixes

DanielHamiltonHOD:
NOTE: Individual teams will have to merge the changes in this PR in to their own branches!

* Maintain Drone deployment step & GitHub Actions for `bluetargeting_dev` in master branch. Ensure UI tests run for changes on `bluetargeting_dev` branch. 
* Deploying `bluetargeting-dev` now requires explicit promotion from Drone UI.
* Adds environment variables for the K8s service URL's for COP Integration Service (`COP_INTEGRATION_SERVICE_URL`) and COP Targeting API (`COP_TARGETING_API_URL`)
* Fixes the certificate request to include short versions of the k8s service name. Removes deprecated `commonName` in favour of `Subject Alternative Name`. Uses `ipAddresses` to ensure IP addresses are listed with type `IP Address` under `Subject Alternative Name`.
* Adds `coreApplication=cop-ui` label. This will be common across all variants of COP UI e.g. `dev`, `bluetargeting_dev`. The intention is to use this common label in other applications network policy - this should then allow us to configure ingress for all variants using a single label.

EDIT:

This has been reverted. Unfortunately, this is not possible at the moment because of the fact we are deploying with Drone, but building the image with GitHub Actions. We should be able to do this at a later date when we move the image build in to Drone.

* Allow deployment to `dev` from any branch. Change removes the need to merge to master when deploying to `dev` but requires explicit promotion to `dev` from Drone UI.


kube/deployment.yml
```diff
@@ -21,7 +19,7 @@ spec:
     metadata:
       labels:
         name: {{.COP_UI_NAME}}
-        service: {{.COP_UI_NAME}}
+        coreApplication: cop-ui
```
michaelruocco, line 22: is this the more common label you were talking about adding to make it easier to enable comms in and out?

michaelruocco (dismissed):
LGTM

kube/deployment.yml
DanielHamiltonHOD: Yes - that's it! We've already done this for some of the E@B2 deployments. So far it seems to be working well for us!

aliktb (dismissed):
Looks good to me. Just a question on the change of heart with the release strategy? I know most of the apps are configured to auto-deploy to DEV on push to master so is this something that will eventually change for all the apps? @DanielHamiltonHOD

DanielHamiltonHOD:
@aliktb Actually, I've had to revert that change - see the updated comment above. But, yes, longer term this should be the way we are going with all our repo's. Clearly, having to merge to the default branch to deploy is not ideal.

aliktb (approved):
Newest commit to revert deploying from any branch looks good. This is hopefully temporary since there is work to do to revise the docker image building process

michaelruocco (approved):
LGTM

PR #517, opened 2022-09-13
autocomplete styling bugfix

keenan218:
The Autocomplete component, when the dropdown is enabled, has a finger pointer cursor style when the mouse is hovered over it. This should be a standard text input style cursor. 

Jira ticket: [COP-11814](https://support.cop.homeoffice.gov.uk/browse/COP-11814)


gardnersam (approved):
LGTM

PR #518, opened 2022-09-13
Fix missing dots in Go templating syntax used by kd

DanielHamiltonHOD:
Fix missing dots in Go templating syntax used by kd


aliktb (approved):
Looks good. Tried to spot any other similar mistakes but so far so good

PR #520, opened 2022-09-13
EAB-1018 make EAB1 path absolute

gardnersam:
This fixes a bug where after using the back button navigating to EAB1 failed. 

The fix is to make the path to the EAB1 form absolute rather than relative.


Deepak-Bagrawat (approved):
LGTM

PR #523, opened 2022-09-14
EAB-879 Add custom component for ePMS email addresses

gardnersam:
Adds custom EAB2 specific component which contains officers email addresses retrieved from ePMS.

```
{
        "id": "epmsOfficerEmails",
        "fieldId": "epmsOfficerEmails",
       "label": "Emails",
       "type": "epmsOfficerEmails",
       "required": true,
       "agency": "HMRC"
}
```

If the agency property is set it will filter the email addresses to only show them according to these rules:
- If agency = HMRC then only display email addresses with the domain "hmrc.gov.uk"  
- If agency = NCA then only display email addresses with the domain "hmrc.gov.uk"
- If agency != HMRC or NCA then only display email addresses with the domain "homeoffice.gov.uk"

If left blank then all email addresses will be shown


Deepak-Bagrawat (approved):
LGTM

PR #524, opened 2022-09-14
COP 11798 - Case View Backlink - routing fixes

bowerscodes-COP:
reinstate the route for /cases/caseId so that emails previously sent, which contain old links, still take the user where they need to go.


aliktb (approved):
Looks good

PR #525, opened 2022-09-14
COP-12004 - Case View route for email links

bowerscodes-COP:
Added a Route to redirect links from older COP emails to the new location within Case View

[COP-12004](https://support.cop.homeoffice.gov.uk/browse/COP-12004)


aliktb (dismissed):
LGTM

keenan218 (dismissed):
Looks good

PR #526, opened 2022-09-15
Enable local dev cop targeting api

charles474:
# Description

Adds proxy config for proxying requests to dev instance.


Shaks (approved):
lgtm

PR #529, opened 2022-09-15
COP-11290: formKey preference from task info over process definition key

kam-cop:
**Problem:** Currently we cannot display react based sub UI task/form to user within a BPMN process.

**Cause:**  The processDecfinition.key always point to starting react form, thus 'view tasks' always redirect user to first form on the BPMN process.

**Solution:** The task/taskInfo's formKey property represent current User Activity that should be performed for a task instance of a BPMN process. By using it over processDecfinition.key solve this issue.

--------------
The following code checks to see if formKey information is available in taskInfo object return in response to /ui/tasks/{taskId} endpoint and uses it to load react form json; and if missing or empty then it uses processDefinition Key to load the react form. json for given task id/instance.

View Task's React Form (Second form on BPMN):

image

Formbuilder Form (Example):
**For FormBuilder forms:** its working because Task Controller endpoint /ui/tasks/{taskId} return valid formbuilder form json as part of the payload.
image

Camunda BPMN with two forms:
image

## Start Page

For the first form (starting form) on BPMN the 'contexts.data.processContext' is undefined and for this starting page id of the form `form.pages[0].id` is returned.

For 'in-progress' (Save and return) form, the 'contexts.data.processContext[`${form.id}`].formStatus.page' exists and is returned.

For second/subsequent form(s) on BPMN, may have 'contexts.data.processContext[`${form.id}`]' but the props 'formStatus.page' can be undefined and and for this starting page id of the form `form.pages[0].id` is returned.


bowerscodes-COP (approved):
looks good

bowerscodes-COP (approved):
looks good

samuelsheldon:
all looks good to me

PR #530, opened 2022-09-20
EAB-1015 Fix save and return for form-with-task

gardnersam:
Add ability to get the correct page for a user in a 'form-with-task' form type based on if they have started the form and where in the form they have reached.

Implements the following behaviours:

- If a user is starting the form fresh then go to the first page
- If a user is returning to the form and saved on a page with type pre-task-list then take them to that page
- If a user is returning to the form and saved on a page in a task (indicated by lack of pre-task-list type) then take them to the hub

This PR also bumps the form-renderer version.


LewisClark-HO (approved):
LGTM

PR #532, opened 2022-09-21
Cop 10818 capture screen

Shaks:
**Description**
Set-up filtered display based on **Airpax** group code 

User will only see **Airpax** view and options if in an **airpax** group. **Airpax** groups will have a type of 8. 

User will only see **Roro** view and options if in an **roro** group. **Roro** groups will have a type of 5. 

If a logged in user's teams is in both 5 and 8 then they get the dropdown option to choose from Airpax/Roro. Default mode to RoRo


client/src/pages/targets/airpax/TargetsListPage.jsx
```diff
@@ -240,6 +221,10 @@ const TargetsListPage = () => {
     };
   }, [refreshTaskList]);
 
+  useEffect(() => {
+    setGroupCodes(GroupUtil.groupCodes(nonRoleGroups, GROUP_CODE_PREFIX.AIRPAX));
```
charles474, line 225: This will have to be more generic as we now have to deal with RCCU groups, and since we will be reusing the components, this will need to no longer be specific to Airpax or RoRo.

client/src/pages/targets/v1/TargetsListPage.jsx
```diff
@@ -360,8 +347,8 @@ const TargetsListPage = () => {
   }, []);
 
   useEffect(() => {
-    setGroupCodes(createGroupCodeArray());
-    setGroupNames(createGroupNameArray());
+    setGroupCodes(GroupUtil.groupCodes(nonRoleGroups, GROUP_CODE_PREFIX.RORO));
+    setGroupNames(GroupUtil.groupNames(nonRoleGroups, GROUP_CODE_PREFIX.RORO));
```
charles474, line 351: Similar to this comment https://github.com/UKHomeOffice/cop-ui/pull/532/files#r976405259

charles474 (approved):

PR #535, opened 2022-09-21
COP-11595 / Dynamic Confirmation page text

bowerscodes-COP:
add ability to handle custom confirmationMessage and confirmationEmails passed in through the form json, and pass associated tests


client/src/components/form/DisplayForm.jsx
```diff
@@ -81,6 +82,9 @@ const DisplayForm = ({
     },
   } = keycloak;
 
+  // eslint-disable-next-line no-unused-vars
```
aliktb, line 85: why do we need this eslint `no-unused-vars`

client/src/components/form/DisplayForm.jsx
```diff
@@ -492,6 +506,8 @@ DisplayForm.propTypes = {
     PropTypes.shape({
       name: PropTypes.string.isRequired,
       id: PropTypes.string.isRequired,
+      confirmationMessage: PropTypes.string,
```
aliktb, line 509: I might be mistaken but I think this actually belongs in the proptypes object below. I think this top one is formIO and the bottom is react. I say that because formIO uses the components property as an array of objects to define the pages whereas with react forms we have a `pages` property which indicates that the bottom `PropTypes.shape` is for react forms

client/src/pages/forms/FormPageSubmission.jsx
```diff
@@ -1,8 +1,21 @@
...
 
 const FormPageSubmission = ({ businessKey }) => {
   const { t } = useTranslation();
+
+  const { formPanelContext } = useContext(FormPanelContext);
+
+  let confirmationEmail = null;
+  let confirmationMessage = null;
+
+  if (formPanelContext !== null) {
```
aliktb, line 15: what if `formPanelContext` had a `confirmationMessage` property but not a `confirmationEmail`. Would we not expect an error when destructuring the object

client/src/pages/forms/FormPageSaved.jsx
```diff
@@ -16,18 +29,26 @@ const FormPageSaved = ({ businessKey }) => {
...
-          {t('pages.form.cop-confirmation-saved-panel.what-next-heading')}
-        </h2>
-        <p className="govuk-body">
-          {t('pages.form.cop-confirmation-saved-panel.what-next-body-p1')}
-        </p>
-        <p className="govuk-body">
-          {t('pages.form.cop-confirmation-saved-panel.what-next-body-p2')}
-        </p>
+        {confirmationMessage && parse(confirmationMessage)}
+        {confirmationEmail && confirmationMessage && (
```
aliktb, line 33: This doesn't make sense to me. Do we want the same confirmation message to show up when a user "saves & returns later"? I would've thought it would be a generic message or if the ask is to customize the save-and-return panel, we would see a different message to when we are submitting the form no?

aliktb (commented):
Some comments here that I think need addressing

client/src/components/form/DisplayForm.jsx
bowerscodes-COP: because formPanelContext is never called within the file and therefore causes linting errors on push

client/src/components/form/DisplayForm.jsx
bowerscodes-COP: Fixed in incoming push, annotations added

client/src/pages/forms/FormPageSaved.jsx
bowerscodes-COP: That would make sense but I understand that is for future development. This ticket was only to enable the functionality. No requirements were given for FormPageSaved scenarios

client/src/components/form/DisplayForm.jsx
aliktb: well could we not remove the `formPanelContext` in the destructured object then?

client/src/components/form/DisplayForm.jsx
```diff
@@ -302,6 +305,19 @@ const DisplayForm = ({
     fileService.setParentProcessKey(form.name);
   }, [form, setRenderer]);
 
+  useEffect(() => {
+    setFormPanelContext({
+      confirmationMessage: form?.confirmationMessage,
```
aliktb, line 310: I think we can initialize these as `null` no?

aliktb (commented):
One last comment

aliktb (approved):
Approving following recent changes and discussions

PR #537, opened 2022-09-21
COP - 10822 - Airpax target capture screen

charles474:
# Description
This PR just adds actions to the airpax targets list page.

A user can now:
- Confirm receipt for a target
- View details for a target
- Capture a target
- FIlter targets on movement direction (`Inbound`/ `Outbound`)

When viewing details, a user can now click on cancel which will return the User back to the targets list.


Shaks (approved):
lgtm

PR #540, opened 2022-09-23
COP-11798/ Case View Backlink email routes

bowerscodes-COP:
reinstate the route for /cases/caseId to redirect emails containing old links


aliktb (approved):
Looks good

PR #541, opened 2022-09-23
Added agency in useEffect dependency array and string interpolation

Deepak-Bagrawat:


gardnersam (approved):
Looks good, thanks for fixing this

PR #542, opened 2022-09-24
COP-10221 - Accordion component

charles474:
# Description
This PR updates dependencies to bring accordion (used within the person component) inline with prototype .


Shaks (approved):
lgtm

PR #543, opened 2022-09-26
COP-11942: Implmented confirmation of receipt at FE

Shaks:
**Confirm receipt** targets would appear in **Receipted** tab

image


client/src/utils/constants.js
```diff
@@ -30,10 +30,12 @@ export const targetTaskNameLike = 'Action the target:%';
 
 export const TAB_STATUS_NEW = 'new';
 export const TAB_STATUS_CAPTURED = 'captured';
+export const TAB_STATUS_RECEIPTED = 'receipted';
 
 export const TARGET_STATUS_NEW = 'NEW';
```
charles474, line 35: Suggestion rather than a must, if we can group this together within an object

e.g. 

```
export const TASK_STATUS = {
  NEW: 'new',
  CAPTURED: 'captured',
  RECEIPTED: 'receipted',
}
```

Same for the `TARGET_STATUSES` all grouped by `TARGET_STATUS` object.

Shaks: Group them both TASK and TARGET_STASUSES

client/src/pages/targets/__components__/TargetListItemOptions.jsx
```diff
@@ -27,18 +23,28 @@ const TargetListItemOptions = ({
...
-      {((selectedTab === TAB_STATUS_NEW && status === TARGET_STATUS_ACKNOWLEDGED) ||
-        selectedTab === TAB_STATUS_CAPTURED) && (
+      {selectedTab === TASK_STATUS.RECEIPTED && (
+        <Button id="assign" classModifiers="blue">
+          Assign
+        </Button>
+      )}
+      {selectedTab === TASK_STATUS.RECEIPTED && (
+        <Button id="viewDetails" classModifiers="secondary">
+          View details
```
charles474, line 43: Why do we have 2 buttons that do the exact same thing. From the looks of it, you've added a second button and forgotten to attach an onClick handler to it.

Regardless, we don't need two buttons that should do the exact same thing.

charles474 (changes requested):
Also looking through the code, we know which endpoint we need to call to get the number of counts if you're a part of an `Airpax` group. The code you added to the `index.jsx` located at the following path `src/pages/home/index.jsx` on the following lines https://github.com/UKHomeOffice/cop-ui/blob/COP-11942/client/src/pages/home/index.jsx#L173-L202,  I don't fully understand why the code there has not yet been updated considering that we can now call the cop-targeting-api locally and we know the parameters that also need to pass to the api.

Try and remember these unfinished portions of code as it is just a bug waiting to happen if it ever made it into production and a user is purely a member of airpax groups only.

client/src/pages/targets/airpax/TargetsListPage.jsx
```diff
@@ -84,13 +81,14 @@ const TargetsListPage = () => {
     setRefreshTaskList(true);
   };
 
-  const createTargetStatusArray = () => {
-    if (selectedTab === TAB_STATUS_NEW) {
-      return [TARGET_STATUS_NEW, TARGET_STATUS_ACKNOWLEDGED];
-    }
-    return [TARGET_STATUS_CAPTURED];
+  const tabStatusMapper = {
```
charles474, line 84: This simple mapper is doing too much, it's also converting to an array.

it really should just be 

```
const tabStatusMapper = {
    new: TARGET_STATUS.NEW,
    receipted: TARGET_STATUS.RECEIPTED,
    captured: TARGET_STATUS.CAPTURED,
  };
```

The comment below then shows how the mapper can then be used.

client/src/pages/targets/__components__/TargetListItemOptions.jsx
Shaks: You are right. I have a added an OR condition to the button code to take care of TASK_STATUS.RECEIPTED

client/src/pages/targets/airpax/TargetsListPage.jsx
```diff
@@ -84,13 +81,14 @@ const TargetsListPage = () => {
...
+    new: [TARGET_STATUS.NEW],
+    receipted: [TARGET_STATUS.RECEIPTED],
+    captured: [TARGET_STATUS.CAPTURED],
   };
 
+  const createTargetStatusArray = () => tabStatusMapper[selectedTab];
```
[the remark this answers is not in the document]
Shaks: fixed

client/src/pages/targets/airpax/TargetsListPage.jsx
Shaks: fixed

client/src/pages/targets/airpax/TargetsListPage.jsx
```diff
@@ -84,14 +81,14 @@ const TargetsListPage = () => {
     setRefreshTaskList(true);
   };
 
-  const createTargetStatusArray = () => {
-    if (selectedTab === TAB_STATUS_NEW) {
-      return [TARGET_STATUS_NEW, TARGET_STATUS_ACKNOWLEDGED];
-    }
-    return [TARGET_STATUS_CAPTURED];
+  const tabStatusMapper = {
```
charles474, line 84: Apologies, I should have seen this also, This really should live a lot higher up this file than randomly in between code.

client/src/pages/targets/airpax/TargetsListPage.jsx
```diff
@@ -84,14 +81,14 @@ const TargetsListPage = () => {
...
+    new: TARGET_STATUS.NEW,
+    receipted: TARGET_STATUS.RECEIPTED,
+    captured: TARGET_STATUS.CAPTURED,
   };
 
   const getTargetTasks = async () => {
+    console.log('selectedTab ', selectedTab);
```
charles474, line 91: `console.log` not required.

charles474 (commented):
Also apply the tab fix to RoRo V1 for consistency.

client/src/pages/targets/airpax/TargetsListPage.jsx
Shaks: My bad

client/src/pages/targets/routes.js
```diff
@@ -30,10 +30,12 @@ const routes = mount({
       route({
         title: context.t('pages.targets.title'),
         view: (
-          <TargetInfoPage
-            targetId={request.params.targetId}
-            targetData={request.params.targetData}
-          />
+          <TaskSelectedTabProvider>
```
charles474, line 33: You need to revert this change here as this is the original targets list page.

Shaks: The issue got fixed applying this change here too

charles474 (approved):
Thanks for making those changes.

PR #553, opened 2022-10-03
COP-11994: Showing relevant task status tabs based on user roles

Shaks:
Showing relevant task status tabs based on loggedIn user roles.

If loggedIn user is either RCCU(bf-rccu-officer) or Higher Officer(bf-higher-officer), Incoming & Receipted tabs are shown.
If loggedIn user is Intercepting Officer(bf-interceipting-officer), Captured target tab is shown (Default role is Intercepting Officer)


client/src/pages/targets/airpax/TargetsListPage.jsx
```diff
@@ -66,6 +67,27 @@ const TargetsListPage = () => {
     FILTERS: 'filter',
   };
 
+  const TASK_TABS = [
+    {
```
charles474, line 71: You might want to key each object for later use when filtering becomes involved and convert into an object instead of an array.

e.g. 
```
const TASK_TABS = {
  ROLE.RCCU: {
  ...
  }
}
```

This will change how you loop through this object

e.g. 

```
Object.keys(TASK_TABS).forEach((k) => {
   const tab = TASK_TABS[k];
   ... Then do your normal checking here.
})
```

client/src/utils/GroupsContext.jsx
```diff
@@ -50,9 +53,24 @@ export const GroupsContextProvider = ({ children }) => {
     setNonRoleGroupCodes(groupCodes);
   };
 
+  const getLoggedInUserRole = () => {
+    const roleGroups =
```
charles474, line 57: This can actually just be

```
const roleGroups = groups.filter((group) => group.grouptypeid === GROUP_TYPE_ROLE).map((group) => group.code) || [];
```

client/src/utils/GroupsContext.jsx
```diff
@@ -50,9 +53,24 @@ export const GroupsContextProvider = ({ children }) => {
...
+      groups.length > 0
+        ? groups.filter((group) => group.grouptypeid === GROUP_TYPE_ROLE).map((group) => group.code)
+        : [];
+    if (roleGroups?.includes('bf-rccu-officer')) {
```
charles474, line 61: Consider moving this to a constant as `'bf-rccu-officer'` might not be the final name. Same for `'bf-higher-officer'`.

client/src/pages/targets/airpax/TargetsListPage.jsx
```diff
@@ -66,6 +67,27 @@ const TargetsListPage = () => {
...
+      id: TASK_STATUS.NEW,
+      label: 'Incoming',
+      heading: 'Incoming',
+      showWhen: [ROLE.RCCU, ROLE.HIGHER_OFFICER],
```
charles474, line 75: For now allow all users to see all tabs.

charles474 (changes requested):
Some comments to go over but give me a shout if there's any you want to discuss or didn't make sense.

client/src/pages/targets/airpax/TargetsListPage.jsx
Shaks: Remove the filter that were checking what tabs to show based on user role. All user can now see all tabs

client/src/utils/GroupsContext.jsx
Shaks: Moved into constant as ROLE_CODE. We could later think about changing it to more appropriate name if needed

client/src/utils/GroupsContext.jsx
```diff
@@ -1,8 +1,9 @@
+/* eslint-disable no-console */
```
charles474, line 1: This can also be removed.

client/src/utils/GroupsContext.jsx
```diff
@@ -2,30 +2,32 @@ import React, { createContext, useEffect, useState } from 'react';
...
-import { GROUP_MEMBER, GROUP_IDS } from './constants';
+import { GROUP_MEMBER, GROUP_IDS, ROLE, ROLE_CODE } from './constants';
 
 const GROUP_TYPE_ROLE = 2;
 
 export const GroupsContext = createContext({
-  groups: {},
-  setGroups: () => {},
+  groups: [],
+  setGroups: () => [],
```
charles474, line 11: Not entirely sure why this has changed when it didn't really need to change.

charles474 (approved):
Thanks for making those changes.

PR #561, opened 2022-10-05
K8s updates and improvements

DanielHamiltonHOD:
* Updates the ingress apiVersion from v1beta1 to v1
* Removes unnecessary egress rule in network-policy
* Uses 'named port' from deployment manifest in ingress, service and network-policy manifests
* Adds new container security options and makes existing explicit


CarolaSeidl (approved):
LGTM

surya-ho (approved):

PR #562, opened 2022-10-05
Merge master into green-dev to pick up the latest deployment changes from platform

philjlharvey:


DanielHamiltonHOD (approved):
Seems to be working OK for Dev.... so should be OK for Green Dev

PR #572, opened 2022-10-13
Cop 8193/your profile

fisherrj2:
Changes to support display of a 'hub-and-spoke' type form for manage my profile.  Changes include a new state variable for hide_title and a conditional alert panel to confirm change of value of manage my profile fields. See https://support.cop.homeoffice.gov.uk/browse/COP-8193


client/src/components/form/DisplayForm.jsx
```diff
@@ -436,6 +455,16 @@ const DisplayForm = ({
       {renderer === Renderers.REACT && (
         <div className="govuk-grid-row govuk-form-group">
           <div className="govuk-grid-column-two-thirds">
+            {form.type === 'hub-and-spoke' &&
+              form.title &&
+              (window.location.pathname.endsWith('edit-your-profile') ||
+                window.location.pathname.endsWith('/hub')) &&
+              form.type && <h1 className="govuk-heading-l">{form.title}</h1>}
```
aliktb, line 462: Just curious with this, is this to "override" the default nature of not showing the form title since it's a `hub-and-spoke` form. I wonder whether we need the extra checks for `edit-your-profile` etc. Unless you want to make this behaviour unique to this form?

client/src/components/form/DisplayForm.jsx
```diff
@@ -478,8 +512,12 @@ const DisplayForm = ({
                       axiosInstance
                     );
                     onSuccess({ ...data, businessKey });
-                    if (type === 'submit') {
+                    if (type === 'submit' && !`${form.id}`.includes('cop-userProfile')) {
```
aliktb, line 515: same with above point about making this a unique solution for this form?

aliktb (commented):
Just a few comments

client/src/components/form/DisplayForm.jsx
fisherrj2: The trouble is that the title is already being displayed in the form renderer for hub type pages but after removing it many tests broke.  It impacted task list.  The trouble is that I understand other applications use FormRenderer and I was reluctant to remove it from there, but at the same time I have to show it here since the alert panel which is now in DisplayForm needs to go between this title and the main body.  This left me with the need to only show it for edit-your-profile hub page.

client/src/components/form/DisplayForm.jsx
fisherrj2: So on submit the action is different depending on whether it is a user profile type form or something else.  With something else there is navigation, but with user profile there is a state update of changed field values so that the success alert panel is displayed on the hub.

client/src/components/form/DisplayForm.jsx
```diff
@@ -444,11 +473,16 @@ const DisplayForm = ({
               }}
               newPageId={newPage}
               viewOnly={false}
+              hide_title={hideTitle}
               hooks={{
+                onGoingBack: () => {
+                  setChangedFieldName(undefined);
```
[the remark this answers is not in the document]
aliktb: I think the logic would also apply for `undefined` but I think `null` is slightly more elegant (at least I think it is)

client/src/components/form/DisplayForm.jsx
fisherrj2: I have made this check for hub-and-spoke type form for now and can make it more specific in future if needed.

client/src/components/form/DisplayForm.jsx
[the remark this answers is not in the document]
fisherrj2: Yes changed to set to null and works a treat and makes code look cleaner.

client/src/components/form/DisplayForm.jsx
fisherrj2: I have made this check for hub-and-spoke type form for now and can make it more specific in future if needed.

client/src/components/form/DisplayForm.jsx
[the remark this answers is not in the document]
aliktb: yep that looks much nicer now :)

aliktb (approved):
Approving PR following discussions/changes

PR #588, opened 2022-10-24
Feature/eab 885 transform eab2 payload for epms

gardnersam:
Converts the payload received from the EAB2 frontend into the format required by ePMS.

Leaves Lockup and Storage location hardcoded / removed as they are not yet available. 

The main part of this PR is removeNonEpmsFields.js. This is where the data mapping happens. Where ever possible the mapping configuration has been captured in epmsPayloadConfig.js. This has mostly been the case where fields need renaming or deleting. More complex mappings and things like default values are not in the configuration file. Hopefully this will mean any changes in the future can mostly be configuration rather than code updates. 

The majority of the rest of the changes are example inputs and outputs used for testing. They may be a bit overkill, happy to discuss reducing them if others think so. 

I've based the mapping off a combination of the ePMS API https://epmswebapidev.azurewebsites.net/swagger/index.html and information from the BAs and Product Owners.


client/src/utils/Form/removeNonEpmsFields.js
```diff
@@ -0,0 +1,179 @@
...
+  return copy;
+};
+
+/**
+ * Convert the payload the EAB2 frontend produces into the format required for ePMS
+ * @param {} payload
+ * @returns
+ */
+const getPayloadForEpms = (payload) => {
+  const payloadCopy = payload;
```
LewisClark-HO, line 154: I think `payloadCopy` is a confusing name. Any changes made to it are also being applied to the original payload here too. Maybe `payloadReference`?

client/src/utils/Form/removeNonEpmsFields.js
```diff
@@ -0,0 +1,179 @@
...
+  fieldsToRemove,
+  itemFieldsToRemove,
+  fieldsToRename,
+  itemFieldsToRename,
+  refDataMappings,
+} from './epmsPayloadConfig';
+
+/**
+ * Applies the mappings found in epmsPayloadConfig to rename/remove/keep fields
+ * @param {*} entry
```
LewisClark-HO, line 12: Comments in this file seem a little incomplete, some params are missing descriptions and type specifiers.

client/src/utils/Form/removeNonEpmsFields.js
```diff
@@ -0,0 +1,179 @@
...
+ * @param {*} removeArray
+ * @param {*} renameArray
+ * @returns
+ */
+const mapData = (entry, removeArray, renameArray) => {
+  return Object.keys(entry).reduce((acc, key) => {
```
LewisClark-HO, line 18: Obviously this is personal preference but I'm not sure what `acc` is meant to be here? I think standard naming for `reduce` is `res` or `result`. A clearer name would be good though I think.

client/src/utils/Form/removeNonEpmsFields.js
```diff
@@ -0,0 +1,179 @@
...
+  });
+};
+
+/**
+ * Update the item fields that have a more complex mapping then rename/move/delete
+ * @param {} item the entry object to update
+ * @returns updated item
+ */
+const updateItemFields = (item, obiEmail, meta) => {
+  const copy = item;
```
LewisClark-HO, line 86: Again, not sure if it affects implementation but this isn't technically a copy. Any changes made to `copy` here are happening to the original `item` too. Another name might be clearer.

client/src/utils/Form/adjustPayloadForType.js
```diff
@@ -14,17 +15,19 @@ const adjustPayloadForType = (type, formName, payload) => {
...
       submittingUsersEmail: submissionPayload.submittingUsersEmail,
     };
+    submissionPayload = getPayloadForEpms(submissionPayload);
   }
   const jsonPart = {
     type: 'json',
     value: JSON.stringify(submissionPayload),
   };
-  if (type === 'submit' || !id || type === 'submitToEpms') {
+  if (type === 'submit' || !id || type === 'convertForEpms') {
```
mark-roberts-ho, line 30: If having multiple comparisons to same variable how about:

`[ 'submit'', 'convertForEpms'].includes( type) || !id`

client/src/utils/Form/removeNonEpmsFields.js
mark-roberts-ho: `acc` is normally `accumulator`, but a longer name might be good.

client/src/utils/Form/removeNonEpmsFields.js
mark-roberts-ho: If it isn't a `copy` there's no need for this constant at all

client/src/utils/Form/removeNonEpmsFields.js
mark-roberts-ho: If you're modifying an object, there's no need to have the `reference` or `alias` at all, is there?

mark-roberts-ho (commented):
Can you move the exampleXYZ files to a separate sub-directory?

client/src/utils/Form/removeNonEpmsFields.js
gardnersam: I did it to satisfy the linter. Directly updating the payload object makes it complain about 
`Assignment to property of function parameter 'payload'  no-param-reassign`
Is there a better way around that?

gardnersam: I could just disable the warning I suppose

client/src/utils/Form/removeNonEpmsFields.js
gardnersam: Fair point, I've completed them

client/src/utils/Form/removeNonEpmsFields.js
gardnersam: swapped to mappedEntry

client/src/utils/Form/removeNonEpmsFields.js
gardnersam: I did this for the same reason, to satisfy the linter, I think I can basically turn off the linter for this check or have the not really needed copy

LewisClark-HO: Seems like a pretty silly check to me, but it must have been turned on for a reason I suppose? My vote would be turn off that check but whatever works.

client/src/utils/Form/adjustPayloadForType.js
gardnersam: Good shout I've swapped to that

LewisClark-HO (approved):
Looks good. Thanks for making those changes.

PR #591, opened 2022-10-24
WIP - Merge bluetargeting-dev to master

charles474:
For Cerberus v2.0.3 release


Shaks (dismissed):
lgtm

client/src/utils/hooks.js
```diff
@@ -204,7 +204,9 @@ export const useFetchGroups = () => {
       const {
         tokenParsed: { team_id: teamid },
       } = keycloak;
-      const allUserGroups = keycloak.tokenParsed.groups.join(',');
+      const allUserGroups = isArray(keycloak.tokenParsed.groups)
```
aliktb, line 207: Just out of curiosity, why would this not be an array?

charles474: I'm not really sure why it wouldn't be an array to begin with. My only guess is someone being overly cautious but a good point.

charles474:
Reopened in https://github.com/UKHomeOffice/cop-ui/pull/597

PR #599, opened 2022-10-27
EAB-721: Custom Validation (Seal Number)

mark-roberts-ho:
## Description

This allows the use of custom validators and implements a custom validator for seal numbers which calls the API to validate the seal numbers. Unlike the other validators, custom validators operate asynchronously on form submit.


client/src/utils/Form/formHooks.js
```diff
@@ -57,11 +69,65 @@ const onTransitionToEabOne = async (form, payload, submittedBy, navigation, axio
...
+
+  if (validate) {
+    debug('Custom validation');
+    const tests = []; // array of validation promises
+    validate.forEach((component) => {
+      // check form contents for field
+      if (Object.prototype.hasOwnProperty.call(payload, component.fieldId)) {
+        debug(`Validating form component ${component.fieldId}`);
+        // run through validations for this field
+        tests.push(runCustomComponentValidation(payload, component));
```
gardnersam, line 89: `runCustomComponentValidation` signature is (component, value), this looks like it's being passed (value, component). If it isn't could these be better named to line up?

client/src/utils/Form/formHooks.js
```diff
@@ -57,11 +69,65 @@ const onTransitionToEabOne = async (form, payload, submittedBy, navigation, axio
...
+      }
+      // check if form has items...
+      if (payload.items) {
+        // check each item
+        payload.items.forEach((item) => {
+          if (Object.prototype.hasOwnProperty.call(item, component.fieldId)) {
+            // validate the item
+            debug(`validating item component ${component.fieldId}`);
+            tests.push(runCustomComponentValidation(item[component.fieldId], component));
```
gardnersam, line 98: I'm not sure what we are expecting `item[component.fieldId]` to be, I think it will be a single field within the item object which makes sense, but doesn't match the call on 89 which takes the whole payload which will contain many components

gardnersam: Also should `component` be the value of the component here?

client/src/components/form/validate/index.js
```diff
@@ -0,0 +1,73 @@
...
+};
+
+/**
+ * run validation onm a component with a given value
+ * @param component
+ * @param value
+ * @returns {undefined|Promise<Awaited<unknown>[]>}
+ */
+const runCustomComponentValidation = (component, value) => {
+  // We only care when we have a value - if we don't have one but want one, set `required: true`.
```
gardnersam, line 60: I don't think required is getting set to true, out of date comment maybe?

client/src/utils/Form/formHooks.js
```diff
@@ -57,11 +69,65 @@ const onTransitionToEabOne = async (form, payload, submittedBy, navigation, axio
   }
 };
 
+const customValidate = async (form, payload, type) => {
```
gardnersam, line 72: would be good to extend formHooks.test.js to cover this

client/src/utils/Form/formHooks.js
```diff
@@ -57,11 +69,65 @@ const onTransitionToEabOne = async (form, payload, submittedBy, navigation, axio
...
+
+    if (tests?.length) {
+      Promise.all(tests).then((results) => {
+        debug('Test results');
+        debug(results);
+        // transform errors into form errors
+        const errors = [];
+        results.forEach((result) => {
+          if (result) {
+            errors.push({ message: result, component: undefined });
```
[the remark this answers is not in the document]
mark-roberts-ho: Yep error caused by my changing the return value of errors - fixed

client/src/utils/Form/formHooks.js
mark-roberts-ho: Agreed

client/src/utils/Form/formHooks.js
mark-roberts-ho: Call on 89 was incorrect - validation only worked over on things with items which is why this wasn't spotted - thanks

client/src/components/form/validate/index.js
mark-roberts-ho: Based on standard runValidation, edited comment

client/src/utils/Form/formHooks.js
mark-roberts-ho: Agreed looking into this

LewisClark-HO (approved):
Looks good!

PR #606, opened 2022-10-31
EAB-1113 Storage Location and Lockup reference

mark-roberts-ho:
## Description

This adds the special selects for Storage Location and Lockup references.


client/src/components/Eab/StorageLocationSelect.jsx
```diff
@@ -0,0 +1,66 @@
...
+        const items = response?.data;
+        const options = [];
+        items.forEach((item) => {
+          const storageLocations = item.storageLocations.map((location) => {
+            return {
+              value: location.value,
+              label: location.description,
+            };
+          });
+          options.push(...storageLocations);
```
LewisClark-HO, line 28: I'm not sure this is doing what the ticket requests. It looks like here we're making a list of all storage locations. I think the options here should only be the locations that are associated with a selected lockup.

client/src/components/Eab/LockupSelect.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+const LockupSelect = ({ id, fieldId, onChange, port, value, ...attrs }) => {
+  const [lockups, setLockups] = useState([]);
+  const axiosInstance = useAxios();
+
+  useEffect(() => {
+    const fetchLockups = async () => {
+      try {
+        const response = await axiosInstance.get(
+          `/camunda/data-service/station-storage-location/${port}`
```
LewisClark-HO, line 17: Out of curiosity, do you know what we would want `port` to be here? I think it will be an object in `formData`. I'm guessing it expects it to be the `id` of the object or something?

mark-roberts-ho: I think you're right - looking into it

client/src/components/Eab/LockupSelect.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+    {
+      station: {
+        value: 'L3',
+        description: 'Lockup 2',
+      },
+      storageLocations: [],
+    },
+  ];
+
+  it('should show lockups', async () => {
```
LewisClark-HO, line 39: I think it would be nice to have a test to make sure the filtering is working vs returning the whole list maybe?

client/src/components/Eab/StorageLocationSelect.jsx
```diff
@@ -0,0 +1,73 @@
...
+        const api = '/camunda/data-service/station-storage-location';
+        let response = await axiosInstance.get(port ? `${api}/${port}` : api);
+        // get whole list if no data
+        if (!response?.data.length) {
+          response = await axiosInstance.get(api);
+        }
+        const options = [];
+        const items = response?.data;
+        // if a lockup provided, filter items by that lockup
+        const lockups = lockup ? items.filter((l) => l.station.value === lockup) : items;
```
LewisClark-HO, line 25: In the case that `lockup` is provided but has no matching storage locations, I think the whole list of storage locations should be shown. I think as it stands this will result in an empty list.

client/src/components/Eab/StorageLocationSelect.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+        },
+        {
+          value: 'S3',
+          description: 'Storage 3',
+        },
+      ],
+    },
+  ];
+
+  it('should show storage locations', async () => {
```
LewisClark-HO, line 38: Similar to the Lockup Select, might be nice to has a test for the filtering vs showing the whole list

mark-roberts-ho: Have implemented test to do this one

client/src/components/Eab/StorageLocationSelect.jsx
mark-roberts-ho: Have changed this to handle that situation

client/src/components/Eab/LockupSelect.test.jsx
mark-roberts-ho: The list is returned from a mock backend, so no frontend work involved in filtering lockups

I have agreed on test for storage locations and implemented it

client/src/components/Eab/StorageLocationSelect.jsx
LewisClark-HO: This isn't needed - clarified with BA on slack. Sorry for making assumptions!

LewisClark-HO (approved):
Looks good, thanks for adding those tests.

PR #607, opened 2022-11-01
Draft: COP-11940: RCCU and non-RCCU able to receive incoming targets

adam-thorpe:
Added the following:
- Airpax RCCU view
  - Incoming, Receipted, Assigned, Accepted and Complete tabs
  - Requires `bf-rccu-officer` group role
- Airpax HO view
  - Incoming, Receipted, Assigned, Accepted and Complete tabs
  - Requires `bf-higher-officer` group role
- Airpax IO view (renamed from Airpax view)
  - Assigned, Accepted and Complete tabs (unchanged)
  - Does not include the user's `groups` in the pages request
  - Requires `bf-intercepting-officer` group role
  
Not Implemented in this PR:
- Receipt target
- Assign target

Notes:
- Currently the IO view displays ALL targets because it is not filtering by `assignee` yet
- No tests written for the new functionality yet


client/src/pages/targets/utils/CommonUtil.js
```diff
@@ -42,6 +61,10 @@ const getViewByPath = (path) => {
   }
 };
 
+const isViewAirpax = (view) => {
```
Shaks, line 64: This method isViewAirpax is confusing. As far as I know, RCCU, HO and IO are roles that are assigned to currently loggedInUser and based on that we would have to show/hide relevant tabs

adam-thorpe: Based on those roles, the user will see different views hence why I've split out the `VIEW.AIRPAX` into three. This method is just a shorthand that allows the program to know if it should render airpax components or not

client/src/pages/targets/utils/CommonUtil.js
```diff
@@ -42,6 +61,10 @@ const getViewByPath = (path) => {
...
+  return view === VIEW.AIRPAX_IO || view === VIEW.AIRPAX_HO || view === VIEW.AIRPAX_RCCU;
```
charles474, line 65: Alternatively

`return [VIEW.AIRPAX_IO, VIEW.AIRPAX_HO, VIEW.AIRPAX_RCCU].includes(view)`

PR #608, opened 2022-11-01
Cop 9277/your profile

fisherrj2:
Add separate alert in DisplayForm for line manager approval request in Manage My Profile.  This alert will be visible as long as the keycloak approval attribute is not set to 'approved'.  See ticket COP-9277 FE - Update My Line Manager.


client/src/components/alert/AlertPanel.jsx
```diff
@@ -21,34 +21,63 @@ const AlertPanel = ({ fieldName, fieldValue }) => {
     }
   }
 
-  function getAlertHeading(name) {
-    if (name === `${t('pages.form.cop-alert-panel.civil-servant-question')}`) {
+  function getAlertHeading(name, status) {
```
aliktb, line 24: Hadn't realised there were some functions which are not arrow functions. Maybe it's just the application showing its age. Should be fine to keep it as is for now

client/src/components/alert/AlertPanel.jsx
```diff
@@ -21,34 +21,63 @@ const AlertPanel = ({ fieldName, fieldValue }) => {
...
+    }
+    return `${t('pages.form.cop-alert-panel.other-msg', { name, value })}`;
+  }
+
+  function getAlertType(name, status) {
+    if (name === `${t('pages.form.cop-alert-panel.line-manager-field')}`) {
+      if (status === 'expired') {
+        return 'error';
+      }
+      return null;
```
aliktb, line 56: just for my understanding, there is a `success` alert type and an `error` type. Is this `return null` just so it doesn't return `success` if it is not `error` but the first if statement is true?

client/src/components/form/DisplayForm.jsx
```diff
@@ -456,8 +458,18 @@ const DisplayForm = ({
...
             )}
-            {newPage === 'hub' && changedFieldName && changedFieldValue && (
-              <AlertPanel fieldName={changedFieldName} fieldValue={changedFieldValue} />
+            {newPage === 'hub' &&
+              changedFieldName &&
+              changedFieldValue &&
+              !changedFieldName.includes('email') && (
+                <AlertPanel fieldName={changedFieldName} fieldValue={changedFieldValue} />
+              )}
+            {newPage === 'hub' && lineManagerApprovalStatus !== 'approved' && (
```
aliktb, line 467: might be me overthinking but would there ever be a situation where both statements were true and you would see 2 `AlertPanel`s

client/src/components/form/DisplayForm.test.jsx
```diff
@@ -235,10 +263,48 @@ describe('DisplayForm', () => {
...
+      <FormPanelContextProvider>
+        <DisplayForm form={testFormRendererData.formData} submitting handleOnCancel={jest.fn()} />
+      </FormPanelContextProvider>
+    );
+    const formRenderer = wrapper.find(FormRenderer);
+    formRenderer.props().hooks.onRequest();
+    expect(mockOnRequest).toHaveBeenCalledTimes(1);
   });
+
+  // it('should send onSubmit event handler into Form when expected', () => {
```
aliktb, line 298: Is this commented out for a reason?

aliktb (commented):
Few comments

client/src/components/alert/AlertPanel.jsx
fisherrj2: I think this is the developer showing his age and I will convert to arrow functions.

client/src/components/form/DisplayForm.test.jsx
fisherrj2: It has been removed

client/src/components/form/DisplayForm.jsx
fisherrj2: Yes there could be two alert panels if there is a line manager approval waiting and a user then changes something else like grade or location.

client/src/components/alert/AlertPanel.jsx
fisherrj2: I think this is an oddity of the Alert component, which then gets exacerbated by npm lint changing the format of the conditional statement.  There are three types of alert: standard, success and error.  To get the standard alert you should supply a null value to the classModifiers prop, otherwise you need to provide success or error.  I will add a comment to clarify this.

client/src/components/form/DisplayForm.jsx
aliktb: and to confirm, this would be the desired outcome on the ui?

client/src/components/alert/AlertPanel.jsx
aliktb: comment looks good to me

client/src/components/form/DisplayForm.jsx
fisherrj2: Yes it is a requirement

client/src/components/alert/AlertPanel.jsx
aliktb: ah, I assumed this was there before you had worked on it. Lets hope you still don't use `var` for your variable declaration in js

client/src/components/form/DisplayForm.test.jsx
aliktb: I just mean delete the commented out code

aliktb: oh never mind haha. Github up to its old tricks not showing me the updated file there

aliktb (approved):
Approving following review comments

PR #611, opened 2022-11-02
Added name property

Deepak-Bagrawat:
https://support.cop.homeoffice.gov.uk/browse/EAB-937


Deepak-Bagrawat:
changes not needed for this file

PR #613, opened 2022-11-02
COP-12299 - Target info page update

charles474:
This PR is a minor update of the targets info page. This PR also displays the control strategies from the TIS form.


Shaks (approved):
lgtm

PR #618, opened 2022-11-04
log out user if team changed

fisherrj2:
Fix for bug 12300 where the user's team does not update on the homepage after changing team in their profile.  Current behaviour is to log the user out and so this is done here.


aliktb (approved):
Looks good to me

PR #622, opened 2022-11-04
Bug/cop 10720

philjlharvey:
Disable the check login iframe functionality. The iframes were not being displayed due to a 'sameorigin' error, and a bug in the keycloak library resulted in hanging promises during the updateToken call. The checkLoginIframe is only required to be able to check for SSO signout on other devices, which will be picked up anyway when the access token expires therefore there is no risk to this fix.

For further reading on the checkLoginIframe option and the implications of disable it, look here:

[Session Status iframe](https://github.com/keycloak/keycloak-documentation/blob/main/securing_apps/topics/oidc/javascript-adapter.adoc#session-status-iframe)

and

[Implications of disabling checkLoginIframe](https://github.com/keycloak/keycloak-documentation/blob/main/securing_apps/topics/oidc/javascript-adapter.adoc#_modern_browsers)


fisherrj2 (approved):
Looks good and cannot see any issues having read the linked notes

PR #625, opened 2022-11-07
Made lockup and storage location return full data

LewisClark-HO:
### Description 
This PR makes it so that the `epmsLockup` and `epmsStorageLocation` custom components return a full data object rather than just a string value.


gardnersam (approved):

Deepak-Bagrawat (approved):
LGTM

mark-roberts-ho (commented):
Approved

PR #626, opened 2022-11-07
Feature/eab 885 convert more field for epms

gardnersam:
Handles new fields for ePMS including 
- special handling
- station
- storage location

Moves soco to item level


LewisClark-HO (approved):
LGTM

PR #630, opened 2022-11-08
When the access token expires, keycloak.updateToken() is called. If t…

philjlharvey:
When the access token expires, keycloak.updateToken() is called. If the update to refresh the token fails (eg if the SSO token has expired) then the keycloak library attempts to call the onAuthLogout event on line 713 of keycloak.js. If this is not implemented, an error is thrown and it all goes over in a heap. Simply implementing this callback enables the normal processing to continue, and the user is redirected to the login page


fisherrj2 (approved):
Looks good

PR #635, opened 2022-11-10
EAB-721 seal number url

mark-roberts-ho:
# Description

Add a route through the backend to call the COP Integration service


server/src/main/resources/application.yml
```diff
@@ -72,6 +72,10 @@ zuul:
       strip-prefix: true
       path: /users/emails
       url: ${COP_IDENTITY_SERVICE_URL:http://localhost:8081/identityservice}
+    integration-service:
+      strip-prefix: true
+      path: /cop-integration
```
gardnersam, line 77: I'm not too sure but should this be `/cop-integration/**` ? Most of the others are, I'd guess it allows you to hit the rest of the path after `/cop-integration`

https://www.baeldung.com/spring-rest-with-zuul-proxy this makes it look like you'd want the `**`

mark-roberts-ho: Done

gardnersam (approved):
looks good!

PR #646, opened 2022-11-14
EAB-721: api call fix

mark-roberts-ho:
This PR resolves an issue over proxying the API calls by using `useAxios` and `axiosInstance`


LewisClark-HO (approved):
LGTM

PR #651, opened 2022-11-15
Bug/cop 10720

philjlharvey:
- Remove all automatic access token refreshes, as these have the unwanted effect of resetting the SSO idle timeout, thus inactivity is never detected leaving sessions open. Instead, check if the access token needs refreshing only in the axios request interceptor. This means it will only be refreshed when the user is active. If they have previously been inactive for > the idle timeout, then the token request will be rejected and they will be logged out

- Fix tests by implementing the newly used isTokenExpired method in the mock of the keycloak object


fisherrj2 (commented):
Just a couple of points to note:
1. The .gitignore file has a warning at the bottom relating to the new line being removed.  I don't know if this is a problem but just wanted to make you aware.
2. The testing for hooks.js is already below the 90% threshold and a number of the new lines are not covered by testing (33-42 and 51).  Overall the coverage is over 90% but by not raising the test coverage as high as possible on new lines of code we make it harder in the future.

fisherrj2 (approved):
Looks good

PR #652, opened 2022-11-15
COP-12332 - Target info pnc warnings

charles474:


client/src/pages/targets/__fixtures__/Entities/Person/PNC/PncCategoriesMother.fixture.js
```diff
@@ -2,12 +2,41 @@ class PncCategoriesMother {
...
+      drugsSupply: null,
+      drugsTrafficking: null,
+      weaponsPossession: null,
+      weaponsSupply: null,
+      goodsSmuggling: null,
+      peopleSmuggling: null,
+    };
+    return this;
+  }
+
```
Shaks, line 39: Use something like the following function to set all properties in an object to null
Object.keys(obj).forEach(key => {
       obj[key] = null;
});
  }

charles474: Good shout 👍 , completely missed that.

Shaks (approved):
lgtm

PR #654, opened 2022-11-16
EAB-1230: Fixed lockup and storage location not retaining values

LewisClark-HO:
### Description
This PR fixes a bug with the `epmsLockup` and `epmsStorageLocation` custom components. These components were not displaying their values when returning to the page via the back link or CYA page change action.

This PR also makes a small change to `setupProxy.js` to fix an issue when running COP-UI locally.


Deepak-Bagrawat (approved):
LGTM

PR #655, opened 2022-11-16
Cop 12427 event at border

michaelruocco:
implements calls to display event at border form when one has been raised upon submissions of target frontline action


client/src/pages/targets/forms/cop-record-outcome.json
```diff
@@ -38,10 +38,6 @@
               {
                 "label": "Insufficient Resources",
                 "value": "INSUFFICIENT_RESOURCES"
-              },
-              {
-                "label": "Target Withdrawn",
-                "value": "WITHDRAWN"
```
michaelruocco, line 44: this is not related to E@B but needed to be removed as it is no longer applicable and was also broken! it raised a border event if this option was used due to a bug in the BPMN, which is incorrect!

client/src/pages/targets/v2/RecordOutcome.test.jsx
```diff
@@ -152,4 +152,34 @@ describe('RecordOutcome', () => {
...
+
+  it('should show border event form if border event raised when front line action recorded', async () => {
+    const reactNaviMock = jest.requireMock('react-navi');
+    const navigationUrls = [];
+    reactNaviMock.useNavigation = () => {
+      return {
+        navigate: (url) => navigationUrls.push(url),
+      };
+    };
+    localStorage.setItem(LS_KEYS.VIEW, VIEW.AIRPAX);
```
charles474, line 164: For the `VIEW` constant, the options are now 

```
export const VIEW = {
  AIRPAX_IO: 'AIRPAX_IO',
  AIRPAX_HO: 'AIRPAX_HO',
  AIRPAX_RCCU: 'AIRPAX_RCCU',
  NONE: 'RORO',
  RORO: 'RORO',
  RORO_V2: 'RORO_V2',
};
```

`VIEW.AIRPAX_IO` Should do the trick.

charles474 (approved):
LGTM

client/src/pages/targets/v2/RecordOutcome.test.jsx
michaelruocco: Cool, will update now, thanks!

michaelruocco: done

PR #658, opened 2022-11-17
COP-12364 - Updates the control strategy field

charles474:
This PR updates the control strategy field on the UI to line up with the field on the target information sheet.

image


michaelruocco (approved):
LGTM

PR #673, opened 2022-11-25
EAB-1182: Added custom component for unit of measure

LewisClark-HO:
### Description
This PR adds a custom component for Unit of Measure to allow for filtering of valid UoM options.

### Tests
Accompanying unit tests are included.


client/src/components/form/DisplayForm.jsx
```diff
@@ -532,6 +533,9 @@ const DisplayForm = ({
                   }
                 },
                 onGetComponent: (config) => {
+                  const activeItem = config.formData?.items?.find(
```
gardnersam, line 536: why not do this within the case?

client/src/components/Eab/UnitOfMeasureSelect.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+  useEffect(() => {
+    const fetchUnits = async () => {
+      try {
+        const api = '/refdata/v2/entities/unit';
+        const response = await axiosInstance.get(api);
+        const validUnits =
+          acceptedUnits.length === 0
+            ? response.data.data
+            : response?.data?.data
```
gardnersam, line 21: how come you have no faith in response existing here but did the line above

client/src/components/Eab/UnitOfMeasureSelect.jsx
```diff
@@ -0,0 +1,81 @@
...
+                ?.map((unit) => {
```
gardnersam, line 22: would a filter make more sense as you are filtering out options rather than transforming a data set?

client/src/components/form/DisplayForm.jsx
LewisClark-HO: The linter doesn't like variable declarations in switch cases :(

client/src/components/Eab/UnitOfMeasureSelect.jsx
LewisClark-HO: Good point. I've now changed this to have renewed faith on line 21.

client/src/components/Eab/UnitOfMeasureSelect.jsx
LewisClark-HO: it absolutely would, thank you

gardnersam (approved):

PR #676, opened 2022-11-28
COP-12325 - Displays passenger status

charles474:
This PR replaces the boarding status with the passenger status from the TIS.


michaelruocco (approved):
LGTM

PR #681, opened 2022-11-29
Improvement to dropdown behaviour on back button action

charles474:
Attempt at improving the dropdown behaviour when the back button is used.


Shaks (approved):
lgtm

PR #688, opened 2022-12-02
COP-12243/new-attribute-for-MMP

keenan218:
Adding temporary change to cop-ui to modify a users Civil Servant status depending on what grade they have assigned to their keycloak profile. This change can be reverted once all users' Civil Servant status has been updated. 

JIRA ticket: [COP-12243](https://support.cop.homeoffice.gov.uk/browse/COP-12243)


client/src/components/form/DisplayForm.jsx
```diff
@@ -86,7 +87,7 @@ const DisplayForm = ({
       realm_access: { roles },
       team_id: teamid,
       session_state: sessionId,
-      civil_servant: areYouACivilServant,
+      civil_servant: areYouACivilServant = getCivilServant(gradeId),
```
fisherrj2, line 90: Maybe change the function name to getCivilServantStatus as that is what it is doing rather than getting an actual civil servant

client/src/components/form/utils/CivilServantHelper.jsx
```diff
@@ -0,0 +1,11 @@
+const getCivilServant = (gradeId) => {
```
fisherrj2, line 1: Just wondered if this actually should be in client/src/utils/Form with the other Javascript function helpers?  Also all of those have a file name that matches the function name.

client/src/components/form/utils/CivilServantHelper.jsx
```diff
@@ -0,0 +1,11 @@
...
+  if (gradeId === 'c27cf914-0b97-4c4e-bb2d-2ae4b886baa5') {
```
fisherrj2, line 2: Maybe add a comment to give some background to this as a new developer would be confused seeing this kind of hardcoding in the application.

client/src/components/form/utils/CivilServantHelper.jsx
keenan218: Thats a good shout actually, i hadn't actually noticed these

client/src/components/form/DisplayForm.jsx
fisherrj2: Just wanted to point out the 'default'!

client/src/components/form/utils/CivilServantHelper.jsx
fisherrj2: The other helper functions in that folder have a function definition section preceding the function.

fisherrj2 (approved):

PR #689, opened 2022-12-02
COP-11992: Implement target assign to groups

adam-thorpe:
Features
- Add a new `groups` multiselect to the Assign page
  - Only renders for `RCCU_receipted` targets
- Sends the result as part of the /assign request


client/src/pages/targets/forms/cop-assign.json
```diff
@@ -43,6 +43,44 @@
               ]
             }
           ]
+        },
```
charles474, line 46: Nothing wrong with this approach. Alternatively you could equally convert this file into a `.js`, embed the form json within a function 

```
const form = (showGroupsField) => {
   return {
     email_config,
     ...(showGroupsField && { <--- This bit here is now your condition, rather than actually fudging it into the form data
         groups_config,
       }),
   };
}

export default form;
```

client/src/pages/targets/v2/Assign.jsx
```diff
@@ -129,6 +134,9 @@ const Assign = ({ targetId }) => {
   return (
     <FormRenderer
       {...form}
+      data={{
```
charles474, line 137: ... Then here 

```
    <FormRenderer
      {...form(showGroupsField)}
      hooks={{
        onGetComponent,
        onRequest: (req) => formHooks.onRequest(req, keycloak.token),
        onSubmit: async (_, payload) => {
          await submitAssign(toSubmissionPayload(payload));
        },
        onCancel: async () => {
          await navigation.goBack();
        },
      }}
    />
```

charles474 (approved):
LGTM

client/src/pages/targets/forms/cop-assign.json
adam-thorpe: Ah interesting. I just went with what I knew how to do. Do you reckon this second approach is better/easier to manage?

adam-thorpe:
This can't be merged until the BE change is also merged/deployed so just waiting on that

client/src/pages/targets/forms/cop-assign.json
charles474: It's about the same but if anything, you then have flexibility to play with the form json config before it is rendered.

client/src/pages/targets/forms/cop-assign.json
```diff
@@ -43,6 +43,44 @@
...
+        {
+          "id": "groups",
+          "fieldId": "groups",
+          "show_when": {
+            "field": "showGroupsField",
+            "op": "eq",
+            "value": true
+          },
```
charles474, line 54: This bit would no longer be required if the cop-assign form is a javascript file with a function which returns json.

PR #694, opened 2022-12-05
COP-12406: warnings are now being displayed from PNC details for a person

Shaks:
Cop-12406: COP-12406: warnings are now being displayed from PNC details for a person


client/src/pages/targets/adapters/row/targetRow.js
```diff
@@ -140,7 +135,7 @@ const toRoRoRow = (target, handleRadioButton, menuName, checkedRadioBtnName) =>
         {RisksUtil.category(selector)}
       </td>
       <td role="cell" className="hods-table__cell">
-        {StringUtil.capitalize(RisksUtil.warning.status(selector))}
+        {MovementUtil.warnings(persons) ? 'Yes' : 'No'}
```
charles474, line 138: Why couldn't we just stick a function within StringUtil which converts boolean to string? Rather than having logic like this here, put it inside a function.

e.g.
`StringUtil.boolToString(MovementUtil.warnings(persons))` <-- This would return either `Yes` or `No` based on the value that is returned from `MovementUtil.warnings(persons)`.

client/src/pages/targets/utils/MovementUtil.test.jsx
```diff
@@ -358,4 +358,72 @@ describe('MovementUtil', () => {
...
+      },
+      passengerStatus: 'DEPARTURE_CONFIRMED',
+    };
+
+    const persons = new AirpaxTargetInfoMother().withPerson(PERSON).build();
+    expect(MovementUtil.warnings(persons)).toBeFalsy();
+  });
+
+  it('should return true if PNC details for persons are found', () => {
+    const PERSONS = [
```
charles474, line 387: Why are we not using the PersonMother fixture?

This approach defeats the purpose of having fixture classes.

client/src/pages/targets/utils/PersonUtil.js
```diff
@@ -30,10 +30,14 @@ const getStaffConcerns = (person) => {
...
-  } = person;
-  return details?.types?.filter((item) => !EXCLUDE.includes(item) && !!item) ?? [];
+
+  if (person?.policeNationalComputerCheck?.details) {
+    const {
+      policeNationalComputerCheck: { details },
+    } = person;
+    return details?.types?.filter((item) => !EXCLUDE.includes(item) && !!item) ?? [];
+  }
+  return [];
```
charles474, line 40: This can also been re-written as such

```
const getPncWarningTypes = (person) => {
  const EXCLUDE = ['STAFF_CONCERN', 'POI_CONCERN'];
  const details = person?.policeNationalComputerCheck?.details;
  return details?.types?.filter((item) => !EXCLUDE.includes(item) && !!item) ?? [];
};
```

charles474 (commented):
Few points to go over.

client/src/pages/targets/utils/MovementUtil.jsx
```diff
@@ -136,6 +137,15 @@ const getMode = (informationSheet) => {
   return informationSheet?.movement?.mode || undefined;
 };
 
+const hasWarnings = (persons) => {
+  const warnings =
+    persons.length &&
```
charles474, line 142: Why are we also checking length here? If the persons array was empty `reduce` simply won't do anything.

client/src/pages/targets/utils/MovementUtil.jsx
```diff
@@ -136,6 +137,15 @@ const getMode = (informationSheet) => {
...
+    persons?.reduce((prev, person) => {
```
charles474, line 143: `persons` is never going to be null. In the worst case scenario, it'll just be an empty array.

client/src/pages/targets/utils/MovementUtil.test.jsx
```diff
@@ -358,4 +358,72 @@ describe('MovementUtil', () => {
...
+        found: null,
+        details: null,
+        categories: null,
+      },
+      watchList: {
+        added: true,
+        referenceNumber: 123567,
+      },
+      passengerStatus: 'DEPARTURE_CONFIRMED',
+    };
```
charles474, line 380: Why are we not using the PersonMother fixture?

This approach defeats the purpose of having fixture classes.

client/src/pages/targets/utils/MovementUtil.test.jsx
```diff
@@ -358,4 +358,72 @@ describe('MovementUtil', () => {
...
+          },
+          categories: null,
+        },
+        watchList: {
+          added: true,
+          referenceNumber: 123567,
+        },
+        passengerStatus: 'DEPARTURE_CONFIRMED',
+      },
+    ];
```
charles474, line 426: Why are we not using the PersonMother fixture?

This approach defeats the purpose of having fixture classes.

client/src/pages/targets/utils/MovementUtil.jsx
```diff
@@ -136,6 +137,15 @@ const getMode = (informationSheet) => {
...
+      return [...prev, ...getPncWarningTypes(person)];
+    }, []);
+  return !!warnings?.length;
```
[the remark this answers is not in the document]
Shaks: This make more sense to me than above code for a function to return Yes or No

charles474 (approved):
LGTM

PR #695, opened 2022-12-05
Cop 12003/case view backlink

fisherrj2:
Case view back link takes user back to homepage after going directly to a case from a link.  See https://support.cop.homeoffice.gov.uk/browse/COP-12003


bowerscodes-COP (approved):
Looks good to me!

PR #702, opened 2022-12-07
COP:12575: Add rccuGroups filter to rccu view

adam-thorpe:
Features
- Add some logic to populate the `rccuGroupCodes` filter param
  - This field only gets populated with any RCCU groups that the user belongs to
  - This is for the RCCU view only


client/src/pages/targets/utils/CommonUtil.js
```diff
@@ -152,6 +152,14 @@ const isViewAirpax = (view) => {
   return [VIEW.AIRPAX_IO, VIEW.AIRPAX_HO, VIEW.AIRPAX_RCCU].includes(view);
 };
 
+const isRccuCode = (groupCode) => {
+  return Object.values(RCCU_GROUP_CODES).includes(groupCode);
+};
+
+const getRccuGroupCodes = (groupCodes) => {
```
charles474, line 159: Missing test for this function.

client/src/pages/targets/v2/TargetsListPage.jsx
```diff
@@ -23,7 +23,7 @@ import { CACHED_TAB, GROUP_MEMBER } from '../../../utils/constants';
 import { DEFAULTS, GROUP_DROPDOWN_OPTIONS, LS_KEYS, STATUS } from './constants';
 
 import { CommonUtil, GroupUtil } from '../utils';
-import { getView, setView } from '../utils/CommonUtil';
+import { getView, setView, getRccuGroupCodes } from '../utils/CommonUtil';
```
charles474, line 26: Missing overall test which tests for the addition of the rccu group codes in the filter params.

PR #705, opened 2022-12-09
COP-12543: increment renderer version

fisherrj2:
Increment form renderer version in cop UI following bug fix for Manage My Profile email validation bug.


bowerscodes-COP (approved):
Looks good

PR #706, opened 2022-12-09
EAB-1237: vehicle cop to epms

mark-roberts-ho:


LewisClark-HO (approved):
Looks good, nice rework :)

PR #707, opened 2022-12-09
COP-12094/page-title - update dependency version

fisherrj2:
Update dependency version of Form Renderer in COP UI to provide fix for bug COP-12094.


aliktb (approved):
LGTM

PR #714, opened 2022-12-13
Bug/eab 1352 update epms payload conversion

gardnersam:
Removes the new field `showRemovalConfirmation` from the payload. 

The LeadOfficerEmail field has been swapped to an autocomplete which resulted in getting an object instead of a string. Line added to swap it back to a string. Could not see a way of using the config to do this - 

transform 

```
"leadOfficerEmail": {
        "value": "user@homeoffice.gov.uk",
        "label": "user@homeoffice.gov.uk"
 }, 
```

into 

```
"obiEmail: "user@homeoffice.gov.uk"
```

If I've missed a better way let me know please.


client/src/utils/Form/removeNonEpmsFields.js
```diff
@@ -163,6 +163,7 @@ const updateEntryFields = (entry) => {
     description: `${entry.entryClass.charAt(0).toUpperCase() + entry.entryClass.slice(1)}`,
     value: entry.entryClass.charAt(0).toUpperCase(),
   };
+  entry.obiEmail = entry.obiEmail?.value;
```
mark-roberts-ho, line 166: I  believe a lot (but not all) of this code can be replaced by the transformation rules I wrote to handle vehicles.
Is it necessary to have some 'real' code here?

mark-roberts-ho (approved):
Approved with one comment

PR #725, opened 2022-12-16
COP-10320: ARIA hidden element must not contain focusable elements

Shaks:
ARIA hidden element must not contain focusable elements


charles474 (approved):
LGTM

PR #731, opened 2022-12-16
COP-10396 - Screen reader blank cells

charles474:
# Description
Fixes a bug where a screen reader software would skip an empty cell in a table instead of audiablly saying `Blank`| `Null`.


michaelruocco (approved):
LGTM

PR #735, opened 2022-12-20
increment renderer version

fisherrj2:
Form renderer version incremented as part of a fix for COP-12094 to change the page title when an error occurs.


keenan218 (approved):
LGTM

PR #738, opened 2022-12-22
EAB-990: add custom submit page

mark-roberts-ho:
Allows the use of a custom submit page


client/src/pages/forms/FormPageSubmission.jsx
```diff
@@ -13,37 +42,27 @@ const FormPageSubmission = ({ businessKey }) => {
     return null;
   }
 
-  const { confirmationEmail, confirmationMessage } = formPanelContext;
+  const { confirmationEmail, confirmationMessage, confirm } = formPanelContext;
+
+  const title = confirm?.title ?? t('pages.form.cop-confirmation-panel.title');
+
+  const body = generateConfirmationMessage(
+    confirm?.message ?? confirmationMessage,
```
gardnersam, line 50: is the confirm.message prop kind of a duplicate of the confirmationMessage one? Do we need both?

client/src/pages/forms/FormPageSubmission.jsx
```diff
@@ -4,6 +4,35 @@ import parse from 'html-react-parser';
 import { useContext } from 'react';
 import { FormPanelContext } from '../../components/form/utils/FormPanelContext';
 
+const generateConfirmationMessage = (message, email) => {
+  const { t } = useTranslation();
+
+  return message ? (
```
gardnersam, line 10: would it be worth adding to FormPageSubmisssion.test.jsx to cover this extra functionality? Looks like it's relatively easy to get at this code there

client/src/pages/forms/FormPageSubmission.jsx
gardnersam: If we didnt could we just extract ` confirmationTitle: form?.confirmationTitle,`

gardnersam (commented):
Few thoughts / questions, looks great overall

client/src/pages/forms/FormPageSubmission.jsx
mark-roberts-ho: The confirm block is a replacement confirmationMessage prop but the latter has been retained for back compatibility

client/src/pages/forms/FormPageSubmission.jsx
mark-roberts-ho: Okat

gardnersam: Fair enough, looks like the existing tests will cover rendering a the message input here anyway

gardnersam: If we needed to boost test coverage later we could probably more of less copy paste the `'renders a custom confirmation message AND the email message when passed confirmationMessage AND confirmationEmail props' test and set confirmationMessage to an array, I think that would be more or less 100% coverage. I'll approve for now anyway

gardnersam (approved):
Looks good, thanks for the answers to my questions

PR #739, opened 2022-12-23
COP-12748 - Handles dropdown switching appropriately

charles474:
- Handles view switching appropriately and default the targets list to RoRo (V2) when a user is a member of RoRo or a member of RoRo & Airpax.
- Renamed drop down options to reflect which version a user is looking at.

If a user has only RoRo groups, the dropdown will contain on my two options:
- RoRo
- RoRo (V1)

If a user has only Airpax groups, the dropdown will not be displayed.

If a user has both Airpax and RoRo groups, the dropdown will contain 3 options:
- RoRo
- Airpax
- RoRo (V1)


Shaks (approved):
lgtm

PR #742, opened 2022-12-24
COP-12791 - Dashboard targets count

charles474:
This PR fixes the previous logic of displaying the number of targets on the dashboard per logged in user.

E.g.

**`Scenario A:`** If a user is a member of RoRo Groups only, they will be shown the number of targets in the incoming tab on the dashboard.
**`Scenario B:`** If a user is a member of Airpax Groups only, depending on who the logged in user is e.g. `RCCU` | 
 `HIGHER_OFFICER` | `INTERCEPTING_OFFICER`, they will be shown the number of targets in either the `RCCU INCOMING` | `HIGH_OFFICER INCOMING` | `INTERCEPTING_OFFICER ASSIGNED` tab on the dashboard.
**`Scenario C:`** If a user happens to be a member of both groups, then **`Scenario A`** takes precedence as the user is automatically defaulted to the RoRo targets list page by design.

## Supplemental (bug fixes)
- This PR closes a loophole whereby a user who is not part of a particular group or in both groups could access the targets list page of another group e.g. an airpax user who is not part of any roro groups could access the roro targets list page via manually altering the URL, vice versa.
- This PR closes a loophole whereby a user who is not part of a particular group or in both groups could access the target info page of another group e.g. an airpax user who is not part of any roro groups could access the roro target info page via manually altering the URL, vice versa.


client/src/pages/home/utils/adaptFilterParams.js
```diff
@@ -0,0 +1,95 @@
+import {
```
charles474, line 1: Still debating if I want to use a config for the different group members however, groups members of **`RORO`** & **`BOTH`** will have identical configs attributes.

michaelruocco (approved):
LGTM

PR #743, opened 2023-01-02
Improvement - Target Info page

charles474:
This PR closes a loophole whereby a user can replace the target Id located in the roro target info page URL with an airpax target Id and the page then attempts to render the task data using the correct components but on an incorrect URL.

In the event this ever happens, the user will just now be redirected to the correct URL and the page renders with the correct components.

# Supplemental
- Resolves a bug where when hovering over a passport photo on the target info page, the whole accordion Item shifts slightly to the right (so far seems to happen locally but this changes in this PR should resolves that behaviour so that it behaves the same as it does when deployed).
- Renders the system checks component when there are system checks otherwise, this whole section is hidden.
- Replaces an overly greedy regex that was more susceptible to ReDOS with an alternative.


michaelruocco (approved):
LGTM

PR #750, opened 2023-01-05
COP-12801 - Removes RoRo v1 targets list page

charles474:
## Description
This PR removes the RoRo V1 targets capture screen, including any links to it.

## Note
- As most users will be overly familiar with the usual URL e.g. `.../targets` that takes a user directly to the **V1** targets capture screen, a user may out of habit enter that URL, have it bookmarked or their browser may even suggest it when they type in the address bar. Should this happen, the user will now simply be redirected to the **New** RoRo target capture screen.

- All additional config which conditionally enabled the new RoRo target capture screen has also been removed.

# To Test
- Pull & run branch against dev.
- Also test the redirect when a user tried to access the now removed v1 target capture screen e.g. `http://localhost:3000/targets` theh should now be redirected to the the new roro targets list page.


michaelruocco (approved):
LGTM

PR #753, opened 2023-01-09
Fixes typo

charles474:


client/public/locales/en-GB/translation.json
```diff
@@ -485,7 +485,7 @@
             "prefix": "You can not record an outcome for target",
             "suffix": "Target is currently assigned to"
           },
-          "not-accepted": "You can not record an outcome against an this target"
+          "not-accepted": "You can not record an outcome against this target"
```
domvarndell, line 488: `cannot` over `can not`? Both are valid but my brain always reads the latter as the former haha

charles474: 😁  good point, I'll update all instances.

domvarndell (approved):
LGTM

PR #728, opened 2022-12-16
COP-12675/eslint forbidden PropTypes errors

bowerscodes-COP:
Replaced instances of PropTypes.object (which is no longer recommended) with PropTypes.shape(), in order to address eslint linting errors. [Jira Ticket](https://support.cop.homeoffice.gov.uk/browse/COP-12675)

PR #754, opened 2023-01-09
Release 2.0.5

charles474:


client/src/pages/home/index.jsx
```diff
@@ -34,26 +29,25 @@ const Home = () => {
...
   const isMounted = useIsMounted();
+  const source = axios.CancelToken.source();
   const { trackPageView } = useMatomo();
   const { groups, nonRoleGroups, nonRoleGroupCodes, groupMember } = useContext(GroupsContext);
+  const { currentGroup, setCurrentGroup, groupLoaded } = useContext(CurrentGroupContext);
+  const [groupChanging, setGroupChanging] = useState(false);
+  const [groupTasksCount, setGroupTasksCount] = useState({ isLoading: true, count: 0 });
+  const [targetTasksCount, setTargetTasksCount] = useState({ isLoading: true, count: 0 });
+  const [yourTasksCount, setYourTasksCount] = useState({ isLoading: true, count: 0 });
+  const currentUser = keycloak.tokenParsed.email;
```
gardnersam, line 42: seems like this only gets used once, is it worth setting it up the top here?

client/src/components/Tabs/Tabs.jsx
```diff
@@ -72,7 +72,7 @@ const Tabs = ({ id, title, items, onTabClick, setActivePage }) => {
 Tabs.propTypes = {
   id: PropTypes.string.isRequired,
   title: PropTypes.string,
-  items: PropTypes.arrayOf(PropTypes.object).isRequired,
+  items: PropTypes.arrayOf(PropTypes.shape()).isRequired,
```
gardnersam, line 75: whats the reason for swapping all these from object to shape?

gardnersam (commented):
Can't see any problems for blue-bfid that stand out, only two comments that don't necessarily require a change, just thoughts/questions

client/src/components/Tabs/Tabs.jsx
charles474: `PropTypes.arrayOf(PropTypes.object).isRequired` was causing the green-team linting issues https://github.com/UKHomeOffice/cop-ui/pull/728. Oddly enough, it wasn't causing us any issues.

client/src/pages/home/index.jsx
charles474: That's a good shout. It doesn't really get used anywhere else.

Missed that one 😬

